PR #65, opened 2018-05-09
Update the migration guide with mandate changes

Smitsel:


Smitsel:
Removed the customerId bit.
Left the mandate changes there.

Issue #46, opened 2018-05-05
Improve legibility


mollierobbert:
The old docs (https://www.mollie.com/en/docs) are far more legible than the new ones.

The content column as well as the default font size are too small.

adriaanmol:
What are you guys going to change then? Please also discuss with Giel.

@mattijsbliek doesn't need to be constituent with payment method pages because we are likely going to change that also another moment (iterating).

mollierobbert:
We want to increase the font size and the width of the body. Additionally the sidebar is currently misaligned. These are your suggestions from our most recent discussion on the new docs styling, Adriaan.

CC @gielcobben

vernondegoede:
- The body font size has been increased a few days ago by @Marinolinderhof (in #71).
- Regarding the width of the content area; the optimal line length for your body text is [considered to be 50-60 characters per line](https://baymard.com/blog/line-length-readability), including spaces. This is currently the case.
- Collapsed subnavigation bar now shows text instead of icons.

mollierobbert:
What about the misalignment of the sidebar?

Issue #57, opened 2018-05-07
In-page header anchor tags don't work


mollierobbert:
When clicking the anchor icon next to a heading, the URL doesn't change.

Same goes for when clicking a submenu item in the sidebar which refers to an in-page header.

mollierobbert:
Fixed in #71.

PR #71, opened 2018-05-09
Minor css js fixes 

Marinolinderhof:
This solves the following:
- https://github.com/mollie/api-documentation/issues/57 fixed url changes whenever you click on the anchor header
- Changed height of github svg with 1px so it aligns better
- changed dash/hypen in submenu items to align in the middle

### DIFF
--- a/source/theme/js/sidebar.js
+++ b/source/theme/js/sidebar.js
@@ -46,7 +46,6 @@ const scrollToSection = event => {
     },
   });
 
-  event.preventDefault();
 };
 
 const toggleSidebarGroup = event => {
--- a/source/theme/navigation.html
+++ b/source/theme/navigation.html
@@ -196,7 +196,7 @@
     </li>
     <li class="sub-navigation-item sub-navigation__item js-sub-navigation-item sub-navigation-item--github">
       <a href="https://github.com/mollie/api-documentation/tree/master/source/{{ sourcename|e }}" class="sub-navigation-item__link" target="_blank" rel="noopener nofollow">
-        <svg width="24px" height="23px" viewBox="0 0 24 23" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" class="sub-navigation-item__icon">
+        <svg width="24px" height="23px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" class="sub-navigation-item__icon">
           <g id="Design" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
             <g id="Desktop:-default" transform="translate(-1054.000000, -71.000000)" fill="#999999" fill-rule="nonzero">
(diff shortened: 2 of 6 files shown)


source/theme/navigation.html
```diff
@@ -196,7 +196,7 @@
     </li>
     <li class="sub-navigation-item sub-navigation__item js-sub-navigation-item sub-navigation-item--github">
       <a href="https://github.com/mollie/api-documentation/tree/master/source/{{ sourcename|e }}" class="sub-navigation-item__link" target="_blank" rel="noopener nofollow">
-        <svg width="24px" height="23px" viewBox="0 0 24 23" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" class="sub-navigation-item__icon">
+        <svg width="24px" height="23px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" class="sub-navigation-item__icon">
```
mattijsbliek: I don’t think accomplishes what you intended, it’s (still?) off a bit:

image

source/theme/styles/components/_sidebar.scss
```diff
@@ -146,8 +146,7 @@
   background: #666666;
   position: absolute;
   left: 0;
-  top: 4px;
-  transform: translateY(-50%);
+  top: 8px;
```
mattijsbliek: Wouldn’t it be more robust to use `top: 50%; transform: translateY(-50%)`?

source/theme/js/sidebar.js
```diff
@@ -46,7 +46,6 @@ const scrollToSection = event => {
     },
   });
 
-  event.preventDefault();
```
mattijsbliek: I don’t really understand how this works, if you don’t prevent the default it should just skip to the anchor tag immediately, right?

source/theme/styles/components/_sidebar.scss
Marinolinderhof: Wouldn't do this. because the line is 1px it becomes unsharp tries to be a half piuxel.

screen shot 2018-05-11 at 11 02 18
vs:
screen shot 2018-05-11 at 11 02 27 

It's not robust either cause when you have a multiline menu entry it will be somewhere in the middle:
screen shot 2018-05-11 at 11 05 32

mattijsbliek: Good point 👍

mattijsbliek:
Would you mind also increasing the font-size to 19px? That would close #46 :)

source/theme/js/sidebar.js
Marinolinderhof: works for me. On build an dev. So could you check again? 

`event.preventDefault();` won't change the url because it prevents it. so after it does it's JS magic it stops. But it should change the URL.

source/theme/navigation.html
Marinolinderhof: Ah you're right `height="23px` should change too.

PR #75, opened 2018-05-13
Add a version switcher for all endpoints that have a v1 and a v2

mollierobbert:


source/theme/api-version-switcher.html
```diff
@@ -0,0 +1,35 @@
+{% if pagename.startswith("reference/v") -%}
+	{%- set is_v1 = pagename.startswith("reference/v1/") -%}
+	{%- set is_v2 = pagename.startswith("reference/v2/") -%}
+	{%- set pagename_v1 = pagename.replace("/v2/", "/v1/") -%}
+	{%- set pagename_v2 = pagename.replace("/v1/", "/v2/") -%}
+	{%- set has_v1 = hasdoc(pagename_v1) -%}
+	{%- set has_v2 = hasdoc(pagename_v2) -%}
+
+	{% if (is_v1 and has_v2) or (is_v2 and has_v1) -%}
```
dabroek: Would've been nice if you'd solved it in a more future-proof fashion: `has_other_version`.

mollierobbert: I did give that some thought, but couldn't figure out an easy way to program it with the limited capabilities of Jinja.

mattijsbliek:
I would prefer if the version switch was positioned in the top left corner so it’s located with the regular navigation and doesn’t eat into the space for the page title.

image

source/theme/api-version-switcher.html
```diff
@@ -0,0 +1,35 @@
...
+		<div class="api-version-switcher">
+			API version:
+
+			<div class="link-dropdown link-dropdown--rtl" data-enhancer="link-dropdown">
```
mattijsbliek: Trying to recreate a native select comes with a whole bunch of accessibility challenges. My advice would be to just use a native select element. If we really want to go custom we can use something like [Dropkick.js](https://github.com/Robdel12/DropKick), which contains the right ARIA attributes etc. But I don’t think it’s worth it for just this one select.

source/theme/styles/_variables.scss
```diff
@@ -74,6 +74,7 @@ $line-height-default: 1.5;
 
 $font-weight-light: 300;
 $font-weight-medium: 500;
+$font-weight-bold: 700;
```
mattijsbliek: You can just use `font-weight: bold` right?

source/theme/api-version-switcher.html
willemstuursma: 👍

mollierobbert: Using Dropkick now, but I'm pretty sure I didn't install it 'the NPM way'.

source/theme/styles/_variables.scss
mollierobbert: Right.

mollierobbert:
Moved the dropdown to the sidebar.

willemstuursma:
Why is this dropkick dependency in VCS and not npm installed?

mollierobbert:
@willemstuursma Because I don't know how it works.

dabroek:
## npm Install
You can install DropKick.js using npm:

`npm install dropkickjs --save`

-- https://github.com/robdel12/DropKick

mollierobbert:
Fixed requested changes @mattijsbliek. Could you generate the lock file for me? My machine is not willing to.

mattijsbliek:
Done! I also improved the non-JS styling of the select, so the transition when JS kicks in is very subtle. Ship it 👍

mollierobbert:
Thanks @mattijsbliek.

PR #90, opened 2018-05-24
Upgrade parcel-bundler dependency

vernondegoede:
Fixes the `eval` warnings with our current CSP.

More information: https://github.com/parcel-bundler/parcel/blob/405bae0d10adf1f05008a44c84530f5bfe2b52c3/changelog.md#fixed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -168,9 +168,9 @@
       "dev": true
     },
     "atob": {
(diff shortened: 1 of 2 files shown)


martijnpolder:
No, this fixes string eval(), which we would frown upon even without CSP.

PR #92, opened 2018-05-28
Add newly available locales for the payment screen

dabroek:


ndijkstra:
The locales are also on this pages: 
- https://docs.mollie.com/reference/v2/customers-api/create-customer
- https://docs.mollie.com/reference/v2/methods-api/list-methods

willemstuursma:
Are there also variants created for the v1 API?

willemstuursma:
Please add it to the changelog too (in that branch).

dabroek:
@ndijkstra The locales for the `list-methods` endpoint differ from the ones in the `payments` and `customer` endpoints.

dmvdbrugge:
I think that while the `list-methods` locales serve a semi-different purpose, the locales can be added there as well. There won't have been done any sorting, but the methods will have been translated.

ndijkstra:
no_NO should be nb_NO: 
https://github.com/TakahikoKawasaki/nv-i18n/issues/9
https://github.com/fzaninotto/Faker/issues/854

dmvdbrugge:
@ndijkstra that's an internal discussion I think. As it stands, Mollie accepts no_NO, so to note otherwise in the documentation is just plain wrong.

dabroek:
@ndijkstra Thank you for bringing this to our attention. I've replaced the invalid `no_NO` with `nb_NO`.

dabroek:
@dmvdbrugge I've added the new locales to the list-methods endpoint to allow the payment methods to be translated and ordered by relevance.

PR #93, opened 2018-05-28
Fix API version switcher on Changelog.

vernondegoede:


willemstuursma:
This could be more future proof by just testing against `v1` / `v2`.

vernondegoede:
@willemstuursma Some filenames include both of these (https://docs.mollie.com/migrating-v1-to-v2). Although this page (currently) doesn't have the API version switcher, being more explicit is better I think.

vernondegoede:
@willemstuursma You're right. Done ✔️

PR #95, opened 2018-05-29
Minor CSS fix: correctly position headerlink.

vernondegoede:
Current situation:

image

New situation:

image


vernondegoede:
:shipit:

PR #98, opened 2018-05-30
Clarify which methods are returned in the List Methods API

willemstuursma:


willemstuursma:
@davdebcom do you think this clarifies it?

PR #101, opened 2018-06-06
Remove device detection

JorikMollie:
As we no longer do this.


willemstuursma:
It's not a sentence @JorikMollie

PR #110, opened 2018-06-15
`amountRemaining` is of type 'amount object'

vernondegoede:
`amountRemaining` should have the same data type as `amountRefunded` and `amount`.


willemstuursma:
Thanks.

PR #112, opened 2018-06-17
Add ESLint + Prettier

vernondegoede:
- Added Prettier dependency which uses Mollie's default configuration file.
- Added ESLint and extend the config from [eslint-config-mollie-node](https://github.com/mollie/eslint-config-mollie/tree/master/packages/eslint-config-node).
- ESLint + Prettier will run against all changed JS files in the `source/theme` folder.
- Ran Prettier against all JS files in `source/theme`.

After running the linters, I fixed the following stuff:
- Added missing `lodash` dependency.
- Ignore `console.log` errors in `toggleClass.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,10 @@
   "version": "1.0.0",
   "description": "API Documentation for all of Mollie's public APIs.",
   "main": "index.js",
-  "scripts": {},
+  "scripts": {
+    "lint:js": "eslint \"./source/theme/**/*.+(js|jsx)\"",
+    "precommit": "lint-staged"
+  },
   "repository": {
     "type": "git",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -3,7 +3,10 @@
   "version": "1.0.0",
   "description": "API Documentation for all of Mollie's public APIs.",
   "main": "index.js",
-  "scripts": {},
+  "scripts": {
+    "lint:js": "eslint \"./source/theme/**/*.+(js)\"",
```
Marinolinderhof: you could change it to "./source/theme/**/*.js\"

Marinolinderhof: should this be added in the make file? `make lint-js`

PR #118, opened 2018-06-27
Upload static assets to other directory to serve on alternate CDN.

hvalle:
We will upload static content to a specific path so that we can point the new CDN (Cloudfront) to this specific path.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -24,3 +24,7 @@ aws s3 cp build/html s3://${AWS_BUCKET}/ --recursive ${AWS_OPTIONS} \
 # Upload static assets
 aws s3 cp build/html/_images s3://${AWS_BUCKET}/_images/ --recursive ${AWS_OPTIONS}
 aws s3 cp build/html/_static s3://${AWS_BUCKET}/_static/ --recursive ${AWS_OPTIONS}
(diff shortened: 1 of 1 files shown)


martijnpolder:
Do you plan on moving the non-asset content to another directory as well to make the assets and html unavailable to each other in the path? I have no security concern if you don't btw.
(in other words, docs.mollie.com/assets/* would serve the assets on the mail domain)

hvalle:
I am not planning on doing it, since it's less changes. If we see the need later, we can do it 👍

deploy.sh
```diff
@@ -24,3 +24,7 @@ aws s3 cp build/html s3://${AWS_BUCKET}/ --recursive ${AWS_OPTIONS} \
 # Upload static assets
 aws s3 cp build/html/_images s3://${AWS_BUCKET}/_images/ --recursive ${AWS_OPTIONS}
 aws s3 cp build/html/_static s3://${AWS_BUCKET}/_static/ --recursive ${AWS_OPTIONS}
+
+# Upload static conetent to assets directory
```
mollierobbert: Typo: 'content' instead of 'conetent'

PR #119, opened 2018-06-28
Load GTM from external file.

vernondegoede:
We need to move our GTM snippet from an external JS file, allowing us to host our assets on a different subdomain. This will fix the CSP issues we're currently experiencing with our GTM snippet.

**Changes**
- Added `gtm.js` to our theme's JS directory.
- When we're generating the HTML for our documentation, we:
  - Copy the `gtm.js` file to the directory where all static files are belonging
  - Generate the HTML using the `sphinx` command 
  - Add the `<script />` tag to all of our HTML files in the `build/html` directory. We're doing this as a build step to prevent this script from being loaded locally.

**Design decisions**
I decided to just copy the `gtm.js` file from `source/theme/js/gtm.js` to `source/_static/gtm.js` because I prefer to have one single place for all JS files. The bundled versions of the JS files will end up in `source/_static` so this folder is not in version control. I could have just put the `gtm.js` script in this folder directly, but I'm afraid in the end it won't be clear where it's located.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,9 @@ source/_static/style.css: source/theme/styles/main.scss node_modules/.bin/parcel
 source/_static/index.js: source/theme/js/index.js node_modules/.bin/parcel
 	node_modules/.bin/parcel build source/theme/js/index.js --out-dir source/_static --out-file index --detailed-report
 
+source/_static/gtm.js:
+	cp source/theme/js/gtm.js source/_static/gtm.js
+
 css-reload:
 	@./node_modules/.bin/parcel source/theme/styles/main.scss --out-dir build/_static --out-file style --no-hmr --port 8001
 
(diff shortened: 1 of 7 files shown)


martijnpolder:
LGTM, though we'll have to see how the GTM code will respond to being in a separate file.

mattijsbliek:
> The bundled versions of the JS files will end up in source/_static so this folder is not in version control.

Some of this folder is under version control, which was confusing to me at first. Ideally we would just have the output of build steps outside of the `source` folder. Out of scope for now though.

Can you move GTM below index.js in `source/theme/layout.html`? You can make them both async so GTM doesn’t have to wait for index.js to finish.

vernondegoede:
> Some of this folder is under version control, which was confusing to me at first. Ideally we would just have the output of build steps outside of the source folder. Out of scope for now though.

Agreed, this doesn't make sense. I'll create a separate issue to keep track of this!

vernondegoede:
@martijnpolder I tested this locally and it works (please ignore the first failed request, this is not related to this page):

image

@mattijsbliek Good call. I made both of them `async`.

mattijsbliek:
For bonus points you could add `<link rel="preconnect" href="https://www.googletagmanager.com" crossorigin />` to the ≤head>, but that’s optional IMO.

PR #122, opened 2018-07-03
Use capital EPS instead of eps.

vernondegoede:
We're using capitalized `EPS` on the marketing website instead of lowercase `eps`.

### DIFF
--- a/source/changelog/v1/changelog.rst
+++ b/source/changelog/v1/changelog.rst
@@ -10,7 +10,7 @@ June 2018
 Monday, 25th
 ------------
 
-- Added the new payment methods Giropay (``giropay``) and eps (``eps``). Note that this method may not be available on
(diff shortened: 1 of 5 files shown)


willemstuursma:
So which one is the right one @vernondegoede? What's the motivation behind the capitizalized spelling? The official spelling is all lowercase: https://www.eps-ueberweisung.at/.

vernondegoede:
Our marketing department decided to communicate with the capitalized version because our competitors are using the capitalized version as well. We're already using this version everywhere in our marketing communication, which was approved by @mollierobbert according to Lieke.

mollierobbert:
Let's take this discussion offline.

mollierobbert:
We went with 'EPS' in all caps in all of our marketing communication to increase legibility, which indeed consciously goes against their style guide.

Issue #124, opened 2018-07-04
Mollie-api-php examples


sandervanhooft:
As discussed [here](https://github.com/mollie/mollie-api-php/issues/217) I am going to add some examples for the Mollie-api-php client.

sandervanhooft:
@willemstuursma Where should the php code samples go?

It seems like the curl examples are not kept in this repo's source files ([example](https://github.com/mollie/api-documentation/blob/5109e1cf818db4827e89de4c8cbd788e27ae5994/source/reference/v2/payments-api/create-payment.rst#request)).

sandervanhooft:
Nvm... They are only visible in RAW mode.

sandervanhooft:
I'll put them below the curl examples. Let me know if I should wrap them or something.

sandervanhooft:
This is what it would look like:

```
Example
-------

Request
^^^^^^^
.. code-block:: bash
   :linenos:

   curl -X POST https://api.mollie.com/v2/payments \
       -H "Authorization: Bearer test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM" \
       -d "amount[currency]=EUR" \
       -d "amount[value]=10.00" \
       -d "description=My first payment" \
       -d "redirectUrl=https://webshop.example.org/order/12345/" \
       -d "webhookUrl=https://webshop.example.org/payments/webhook/" \
       -d "metadata={\"order_id\": \"12345\"}"

.. code-block:: php
   :linenos:

    $mollie = new \Mollie\Api\MollieApiClient();
    $mollie->setApiKey("test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM");
    $payment = $mollie->payments->create([
      "amount" => [
          "currency" => "EUR",
          "value" => "10.00" // You must send the correct number of decimals, thus we enforce the use of strings
      ],
      "description" => "My first payment",
      "redirectUrl" => "https://webshop.example.org/order/12345/",
      "webhookUrl" => "https://webshop.example.org/payments/webhook/",
      "metadata" => [
          "order_id" => "12345",
      ],
    ]);
```

sandervanhooft:
I'll leave the response examples untouched

willemstuursma:
@vernondegoede can we work with this?

sandervanhooft:
@willemstuursma @vernondegoede Gentle reminder :)

I'd like to update the docs with PHP client examples, preferably in the coming weeks - let me know in what format to contribute.

[...]

PR #129, opened 2018-07-20
Added php client example for creating a payment

sandervanhooft:
As discussed in #124 . Here's a single example.

If this is the way to go, I can deliver more next week.

screenshot


willemstuursma:
Looks fine to me, what do you think @mattijsbliek ?

PR #134, opened 2018-07-23
Php examples for Payment, Method and (partial) Refund resource

sandervanhooft:
Added php examples for Payment resource:
- cancel
- get
- list

Added php examples for Method resource:
- list
- get (including issuers)

Added php example for Refund resource:
- create

(Struggling a bit with the Refund `get` method, will add later on.)


willemstuursma:
Thanks @sandervanhooft

PR #135, opened 2018-07-24
Minor fixes

justincase:


justincase:
There's also a discrepancy between the Settlement and Invoice fields. The migration guide refers to `amountNet`, `amountVat` and `amountGross` for both resources. However, the names are reversed in the Invoice example response: `netAmount`, `vatAmount`, `grossAmount`,

willemstuursma:
Thanks @justincase.

I'll update the migration guide. Unfortunately we let this inconsistency slip in.

PR #138, opened 2018-07-26
Add partners menu item.

vernondegoede:


mattijsbliek:
Shouldn’t we have Enterprise in the footer?

vernondegoede:
We don't have a footer in our documentation :-)

Issue #153, opened 2018-08-06
Order response: amountCaptured, amountRefunded are optional


sandervanhooft:
At the moment, they are not marked as such.

sandervanhooft:
Relates to this page: https://docs.mollie.com/reference/v2/orders-api/get-order

sandervanhooft:
Same for `locale`.

Also, the documentation seems to be about setting the locale instead of retrieving it.

PR #154, opened 2018-08-08
Order response fixes

ndijkstra:


sandervanhooft:
@ndijkstra 

#153 is probably relevant here.

sandervanhooft:
@ndijkstra Shouldn't the `method` be included in the Order.get response?

sandervanhooft:
Same for `locale`

sandervanhooft:
And `orderNumber`

sandervanhooft:
From the docs:
> Creating an order will automatically create the required payment to allow your customer to pay for the order.

So the `method`, `locale`, `webhookUrl` and `redirectUrl` fields seem to relate to the `Payment` resource that's created(?). Should there be a `paymentId` or complete payment resource included in the response? Is there another way to retrieve the payment?

(Obviously the `checkout` link can be used for finishing the payment, but that's not for retrieving the related payment(s)).

willemstuursma:
@sandervanhooft eventually we'll provide an `/v2/orders/ord_xxx/payments` endpoint.

ndijkstra:
Thanks @sandervanhooft I will add them

willemstuursma:
Thanks @ndijkstra

PR #161, opened 2018-08-14
Use separate CDN to serve assets

hvalle:
This makes use of the CDN (Cloudfront). The assets are now pointing to the absolute address as I did not find any easy way to override the functionality.

Please feel free to give feedback!


willemstuursma:
This doesn't seem like the right way to do it. 

> I did not find any easy way to override the functionality.

Are you _sure_ there is not an easy way. This breaks all kinds of development flows.

hvalle:
I have looked into the he Sphinx [configuration](http://www.sphinx-doc.org/en/master/usage/configuration.html), the [extensions](http://www.sphinx-doc.org/en/master/usage/extensions/#third-party-extensions), the [third-party extensions](https://bitbucket.org/birkenfeld/sphinx-contrib/src/default/), and even the [awesome sphinx](https://github.com/yoloseem/awesome-sphinxdoc#extensions) GitHub repo.

I can only think of creating our own extension to support this behaviour and somehow add a prefix to generated static content.

What do you think?

hvalle:
I will look into doing this as a post Sphinx build process.

vernondegoede:
@hvalle We insert the script tag for Google Tag Manager during the build step. Not sure if something similar could be used for this as well.

It happens in this script: https://github.com/mollie/api-documentation/blob/master/scripts/gtm.js

willemstuursma:
@vernondegoede that script could also be tremendously improved with `sed` instead of hundreds of npm dependencies.

mattijsbliek:
Actually, that script has 4 dependencies. Also, it works and isn’t causing issues. So I would say our time is best spent elsewhere. Happy to review a PR if you want to rewrite it to sed though 🤷‍♂️

willemstuursma:
@mattijsbliek you're right.

hvalle:
I just pushed a change that uses **sed**, i also did the GTM because it was very easy to add. 🎉

Makefile
```diff
@@ -50,8 +50,20 @@ lint-js:
...
 
-write-gtm:
-	node scripts/gtm.js
+sed:
+	make sed-gtm
+	make sed-cdn
+
+# Go thru all the files, and replace the snippet with the google tag manager code, then cleanup.
+sed-gtm:
+	@LC_CTYPE=C LANG=C find build/ -type f -name '*' -exec sed -i.bak 's/<!-- GOOGLE_TAG_MANAGER -->/<script type=\"text\/javascript\" src=\"\/_static\/gtm.js\" async><\/script>/g' {} \;
```
mattijsbliek: I don’t think this will throw when there are no substitutions right? We should probably add a "DO NOT REMOVE" comment to the GTM comment in the layout then. Cause we won’t know if somebody removes it :)

Also, can you remove `scripts/gtm.js` and its dependencies?

hvalle: I think adding the DO NOT REMOVE is going now too much out of scope for this PR. But I've removed the script and it's dependencies now.

willemstuursma:
@hvalle I would just recommend adding the sed commands to the `html-production` step and not as a separate make target.

hvalle:
All changes are in, I'm merging now.

PR #163, opened 2018-08-15
CS-43: Add minimal / maximum app fee

JorikMollie:
Added the minimum and maximum app fee to the documentation as per https://mollie.atlassian.net/browse/CS-43


willemstuursma:
Should it not be 95% instead of 5%? Regardless, we should use `.` as decimal separator and not `,` as the documentation is in US English.

Smitsel:
This was missed by me, ill fix it.

PR #164, opened 2018-08-15
Fix notation of the application fee formula

Smitsel:


willemstuursma:
@Smitsel can you use the multiplication sign (×) (not "x") instead of `*`.

PR #169, opened 2018-08-17
PROC-1227 Add new method params to docs

Smitsel:


Smitsel:
The `billingEmail` is going to be used for Slice it in the future.
Should we keep the description like this, or maybe add that info already?

willemstuursma:
> Should we keep the description like this, or maybe add that info already?

This is fine.

Sorry I committed with the message _Typo_ but that's a mistake, it was an elaboration.

willemstuursma:
@sandervanhooft fyi

willemstuursma:
@davdebcom fyi

PR #172, opened 2018-08-17
PROC-1245 Add guide for order and order line statuses

Daanvm:
Images are temporary and need to be recreated by someone who can make it look nice.


willemstuursma:
Maybe it is good to have a general introduction to explain that there are bascially two flows: the authorized and the paid flow.

willemstuursma:
Only merchants can cancel order lines, customers cannot affect this.

willemstuursma:
Underlining the various states with the next level header (^ or ~ ?) will give them nice anchors

PR #173, opened 2018-08-20
Change statuses to new design

ndijkstra:


source/reference/v2/orders-api/cancel-order.rst
```diff
@@ -15,7 +15,7 @@ The order can only be canceled while the order's ``status`` field is either ``cr
 
 #. In case of ``created``, all order lines will be canceled and the new order status will be ``canceled``.
 #. In case of ``authorized``, the authorization will be released, all order lines will be canceled and the new order
-   status will be ``void``.
+   status will be ``canceled``.
 #. In case of ``shipping``, any order lines that are still ``authorized`` will be canceled. The new order status will be
```
mvdpanne: Order lines that are `shipping` will be completed, as well.

source/reference/v2/shipments-api/get-shipment.rst
```diff
@@ -172,7 +172,7 @@ Response
                 "imageUrl": "https://sh-s7-live-s.legocdn.com/is/image//LEGO/42083_alt1?$main$",
                 "sku": "5702016116977",
                 "type": "physical",
-                "status": "shipped",
+                "status": "shipping",
```
mvdpanne: Might be that this should be `completed`, depending on whether the whole order line has been shipped, or just a part of it.

mvdpanne: Because we don't have any details on the actual order lines, both could be correct, so it's fine to leave it like this.

PR #176, opened 2018-08-22
PROC-1254 Add isCancelable field to the docs

Smitsel:


Smitsel:
Ping @sandervanhooft, isCancelable has been added :)

PR #177, opened 2018-08-22
PROC-1188 List order refunds

willemstuursma:
@shirshir what do you think.


Smitsel:
I like it, but maybe already add the `lines` documentation to the Get Refund reference.

willemstuursma:
OK, I added it

PR #183, opened 2018-08-29
PROC-1316 Update webhook and redirect urls

willemstuursma:
@sandervanhooft

### DIFF
--- a/source/reference/v1/payments-api/create-payment.rst
+++ b/source/reference/v1/payments-api/create-payment.rst
@@ -52,7 +52,7 @@ Parameters
        .. type:: string
           :required: true
 
-     - `The URL the customer will be redirected to after the payment process. It could make sense for the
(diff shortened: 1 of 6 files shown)


sandervanhooft:
👍 Looks good to me

Smitsel:
Why do we add `required` to GET responses?
Should these not be `null`? Because we don't `require` anything on a get response right?

sandervanhooft:
Yes, that `required` label has been confusing to me as well. Perhaps inverse it with an `optional` label for responses?

willemstuursma:
Its a bit tricky in the source yes, `:required: false` causes the text _Optional_ to appear.

willemstuursma:
@mvdpanne ☝️

willemstuursma:
@Smitsel @sandervanhooft there is a (subtle) difference between a key is optional and a key can be `null`.

willemstuursma:
@mvdpanne yeah it's fine for now

PR #184, opened 2018-08-29
Add orders.read and orders.write to OAuth Permissions

vernondegoede:


willemstuursma:
the permission is also used for managing shipments.

vernondegoede:
@willemstuursma Thanks! Fixed.

PR #185, opened 2018-08-29
PROC-1315 -- "at-fields" documentation

ibudasov:


source/reference/v2/orders-api/cancel-order.rst
```diff
@@ -84,6 +84,7 @@ Response
         "isCancelable": false,
         "metadata": null,
         "createdAt": "2018-08-02T09:29:56+00:00",
+        "expiresAt": "2018-08-30T09:29:56+00:00",
```
ibudasov: Please have a look here: seems like this field was missing before. In my opinion it is returned always after creation

source/reference/v2/orders-api/create-order-refund.rst
```diff
@@ -17,7 +17,7 @@ Create order refund
 
 
 When using the Orders API, refunds should be made against the order. When using *pay after delivery* payment methods such
-as Klarna Pay later, this ensures that your customer will receive credit invoices with the correct product information on
+as *Klarna Pay later*, this ensures that your customer will receive credit invoices with the correct product information on
```
ibudasov: All the across repo "Klarna Pay late"r is consistent since now

PR #189, opened 2018-08-31
Add documentation for update order endpoint

mvdpanne:
- Added a response example to the create order endpoint documentation.
- Minor fix to the update subscription endpoint documentation.

### DIFF
--- a/source/reference/v2/orders-api/create-order.rst
+++ b/source/reference/v2/orders-api/create-order.rst
@@ -588,3 +588,139 @@ Request (PHP)
          ]
        ]
    ]);
+
+Response
+^^^^^^^^
+.. code-block:: http
+   :linenos:
+
+   HTTP/1.1 200 OK
+   Content-Type: application/hal+json; charset=utf-8
+
+   {
+       "resource": "order",
+       "id": "ord_pbjz8x",
+       "profileId": "pfl_URR55HPMGx",
+       "method": "klarnapaylater",
+       "amount": {
+           "value": "1027.99",
+           "currency": "EUR"
+       },
+       "amountCaptured": {
+           "value": "0.00",
+           "currency": "EUR"
+       },
+       "amountRefunded": {
+           "value": "0.00",
+           "currency": "EUR"
+       },
+       "status": "created",
+       "isCancelable": true,
+       "metadata": {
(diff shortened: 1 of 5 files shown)


source/reference/v2/orders-api/create-order.rst
```diff
@@ -588,3 +588,139 @@ Request (PHP)
          ]
        ]
    ]);
+
+Response
+^^^^^^^^
```
ibudasov: Don't you want say here something like: ....order object as described in `Get order`.

source/reference/v2/orders-api/update-order.rst
```diff
@@ -0,0 +1,246 @@
...
+   $order->billingAddress->title = "Dhr";
+   $order->billingAddress->givenName = "Adriaan";
+   $order->billingAddress->familyName = "Mol";
+   $order->billingAddress->email = "adriaan@mollie.com";
+   $order->billingAddress->phone = "+31208202070";
+   $order->update();
+
+Response
+^^^^^^^^
+.. code-block:: http
```
ibudasov: same here: isn't just a link simpler?

mvdpanne: I was confused about this at first as well, but it seems that throughout the documentation, we have both a `Response` bit that mentions that links to another page, and a `Response` bit that shows an example response.

So to be clear: that link is actually in the file, but you'll have to fully open the file to be able to see it.

source/reference/v2/orders-api/update-order.rst
```diff
@@ -0,0 +1,246 @@
...
+           "billingAddress": {
+               "streetAndNumber": "Keizersgracht 313",
+               "city": "Amsterdam",
+               "region": "Noord-Holland",
+               "postalCode": "1234AB",
+               "country": "NL",
+               "title": "Dhr",
+               "givenName": "Adriaan",
+               "familyName": "Mol",
+               "email": "adriaan@mollie.com",
```
willemstuursma: Don't do this!

mvdpanne: I more or less copy pasted this from the create order documentation, didn't know it was a problem.

I will fix it in both endpoints.

PR #193, opened 2018-09-03
PROC-1199 Update shipment tracking information documentation

Smitsel:
Added new endpoint for updating the tracking information for a shipment.

Ping @sandervanhooft.

### DIFF
--- a/source/reference/v2/shipments-api/update-shipment.rst
+++ b/source/reference/v2/shipments-api/update-shipment.rst
@@ -1,0 +1,188 @@
+Update shipment
+===================
+.. api-name:: Shipments API
+   :version: 2
+
+.. warning::
+   This API is currently in private beta. If you are interested in participating, please contact your account manager at
+   Mollie.
+
+.. endpoint::
+   :method: PATCH
+   :url: https://api.mollie.com/v2/orders/*orderId*/shipments/*shipmentId*
+
+.. authentication::
+   :api_keys: true
+   :oauth: true
+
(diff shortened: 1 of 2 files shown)


willemstuursma:
@Smitsel there is no error condition info (e.g. if the shipping address cannot be updated).

Smitsel:
But it can always be updated, we don't have restrictions for that and neither does Klarna. 
Or am i missing something?

https://developers.klarna.com/api/#order-management-api-add-shipping-info-to-a-capture

source/reference/v2/shipments-api/update-shipment.rst
```diff
@@ -0,0 +1,188 @@
...
+--------
+``200`` ``application/hal+json; charset=utf-8``
+
+A shipment object is returned, as described in
+:doc:`Get shipment </reference/v2/shipments-api/get-shipment>`.
+
+Example
+-------
+
+Request (curl)
```
mvdpanne: No PHP request example?

mvdpanne:
I don't think any error condition info is necessary. If that call fails, then it's not something the user of our API can fix, and it makes sense to throw a 500.

source/reference/v2/shipments-api/update-shipment.rst
Smitsel: I've just asked @sandervanhooft to create one in the php client repo and here afterwards:
https://github.com/mollie/mollie-api-php/pull/261

Issue #127, opened 2018-07-18
API and API version should be in document title


willemstuursma:
So that I can distinguish the API versions in the Google search results: 
![screen shot 2018-07-18 at 16 30 48](https://user-images.githubusercontent.com/701299/42888194-fdcd410e-8aa7-11e8-9644-46213d689e2e.png)

willemstuursma:
@vernondegoede  could you give this a little nudge if you have some time to spare?

willemstuursma:
@vernondegoede could you help @MollieRick figure this out?

MollieRick:
Fixed in #200

PR #200, opened 2018-09-05
Adding API-version to document title

MollieRick:
See issue #127

### DIFF
--- a/source/theme/layout.html
+++ b/source/theme/layout.html
@@ -1,11 +1,17 @@
 {% set url_root = pathto('', 1) -%}
+{%- set is_v1 = "/v1/" in pagename -%}
+{% if is_v1 %}
+  {%- set versionName = "v1" -%}
+{% else %}
+  {%- set versionName = "v2" -%}
+{% endif %}
 <!DOCTYPE html>
 <html lang="en">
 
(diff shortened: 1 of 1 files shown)


willemstuursma:
This looks good to me! `versionName` is always set.

willemstuursma:
Maybe `Create payment — Mollie v1 API documentation` would be even better.

willemstuursma:
@ndijkstra what do you think?

ndijkstra:
@willemstuursma I like your suggestion `Create payment — Mollie v1 API documentation`

MollieRick:
@willemstuursma @ndijkstra How about `Create payment - Mollie API v1 documentation`? It is version 1 of the API, not Mollie :-)

vernondegoede:
I don't have access to my laptop right now, but how does this look for guides? We don't want to set an API version number for these pages.

MollieRick:
@willemstuursma @vernondegoede Done. Please check again

mvdpanne:
Seems fine to me? Not sure what you mean with magic strings but I'd say lets merge it.

vernondegoede:
LGTM 🚢 it

Issue #201, opened 2018-09-06
Sidebar incomplete for Recurring payments


willemstuursma:
Recurring payments has several headers that are missing in the sidebar.

When clicking "Recurring payments" in the guides I expect all relevant headers to be visbible.

https://docs.mollie.com/payments/recurring

For example, it's missing _Charging immediately on-demand_ and _Charging periodically with subscriptions_. 

Could you have a look at it?

Issue #195, opened 2018-09-03
Subscription vs recurring payments confusion


sandervanhooft:
While working on Laravel Cashier, I notice that the distinction between **Subscriptions** and **Recurring Payments** is not clear.

I am not sure when to choose what, and what the limitations/consequences will be, both short term and long term. When talking to Mollie representatives, I get a different response each time.

Perhaps this can be clarified further in the docs?

At the moment, I am not entirely sure what Subscriptions really offer on top of Recurring Payments.

A few things I noticed:
- subscription can be updated, payment not
- subscriptions offer nice separate dashboard features, so it's easy to see what revenue is recurring
- subscriptions do not offer metadata
- payment offers metadata, but I'm not sure it carries over to subsequent payments
- I'm not sure what happens when Mollie decides to cancel a subscription (webhook?) i.e. after a customer complaint
- retrieving past subscription's payments is complex right now (it's on backlog)
- getting upcoming "payment date" is impossible right now for both payment and subscription (it's on backlog for subscription)

willemstuursma:
#201

willemstuursma:
Quoting @MollieRick: 

> Maybe it's nice to have a little table with the limitations and differences between On-demand and Subscriptions.

PR #202, opened 2018-09-06
Fixing #201 - Adding missing headers to sidebar

MollieRick:


willemstuursma:
Thanks @MollieRick 
Any more headers that are missing? Anything that is unclear or can be improved on this doc page or topic?

MollieRick:
@willemstuursma All headers are now present in de sidebar. I agree that the recurring guide can be a bit confusing (#195) because of the naming and limitations. Maybe it's nice to have a little table with the limitations and differences between On-demand and Subscriptions. My knowledge is not that far at the moment because I'm new, so maybe can take someone else this as a task?

PR #204, opened 2018-09-07
PROC-1355 Change the cancel order line to support multiple line cance…

Smitsel:
We've changed the way we go about canceling a line and now support multiple lines aswel.

This is breaking for the PHP client.
@sandervanhooft can you change it there and change the PHP example accordingly?


willemstuursma:
Looks fine, maybe add some hints that it is similar to the Create Shipment API.

ndijkstra:
@Marvin-Magmodules FYI: We've changed our Cancel order lines endpoint.

PR #218, opened 2018-09-12
Improve PHP syntax highlighting.

### LINKED ISSUE #149: Improve PHP syntax highlighting
The colors that are being used for the PHP syntax highlighting don't match with the colors we use for our `curl` example requests. Let's use the blueish syntax highlighting for all requests.

https://github.com/mollie/api-documentation/blob/master/source/theme/styles/components/_highlight.scss

image

vernondegoede:
Fixes #149.

### DIFF
--- a/source/theme/styles/components/_highlight.scss
+++ b/source/theme/styles/components/_highlight.scss
@@ -43,6 +43,7 @@
 .rst-content .highlight .kc,
 .rst-content .highlight .p,
 .rst-content .highlight .l,
+.rst-content .highlight .mf,
 .rst-content .highlight .s2 {
   color: $grey-dark;
 }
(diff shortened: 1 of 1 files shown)


mollierobbert:
Nice work, looks much better.

sandervanhooft:
@vernondegoede I like how it matches the Mollie theme!

But I'm not sure it's the most balanced in terms of readability, the black is heavier than the blue in my eyes. So the strings get more attention than the operands. ( = string contents get more attention than structure).

It can be me though, not the best set of eyes :).

mollierobbert:
Thanks for the feedback @sandervanhooft, much appreciated. Accessibility is more important to us than design. @vernondegoede can you tweak it further based on this feedback?

PR #222, opened 2018-09-13
Temporary redirect solution. 

vernondegoede:
This is far from ideal, but it should fix the broken pages until we implement the AWS Lambda redirects.

Tested locally.


willemstuursma:
"Migratig"

vernondegoede:
Thanks @willemstuursma. Just noticed your comment. Fixed in #229.

PR #225, opened 2018-09-17
Mollie/proc 1232 orders webhooks guide

Daanvm:


source/guides/webhooks.rst
```diff
@@ -1,5 +1,87 @@
...
+with the ID of the updated object. When Mollie calls your webhook, you should fetch the latest status and process it if
+the status was changed.
+
+Example
+-------
+
+The most important example of a webhook is when a payment is paid. If you created the payment with a webhook URL, we
+will call that webhook URL with a single POST-parameter called ``id`` and a value of for example ``tr_d0b0E3EA3v``. The
+script behind your webhook URL should use that ID to :doc:`fetch the payment status </reference/v2/payments-api/get-payment>`
+and act accordingly. If the new payment status is ``paid`` you can start shipping the order.
```
MollieRick: Maybe mentioning that, because Mollie needs to reach the server, `localhost` is not an option?

Daanvm: Good idea, I've added a link to a helpdesk article just like the link to the IP address article.

PR #227, opened 2018-09-17
Update examples to reflect the docs

lvonk:
According to the docs the fields `createdDatetime` and `settledDateTime`
have been renamed to `createdAt` and `settledAt`.


willemstuursma:
Thanks.

PR #230, opened 2018-09-18
Adding metadata-field to subscriptions

MollieRick:


willemstuursma:
Nice example usage. Don't forget the changelog.

willemstuursma:
@sandervanhooft fyi this is coming up soon.

source/reference/v2/subscriptions-api/create-subscription.rst
```diff
@@ -112,6 +112,15 @@ Replace ``customerId`` in the endpoint URL by the customer's ID, for example
...
 
+   * - ``metadata``
+
+       .. type:: mixed
+          :required: false
+
+     - Provide any data you like, for example a string or a JSON object. We will save the data alongside the
+       subscription. Whenever you fetch the subscription with our API, we'll also include the metadata. You can use up
+       to approximately 1kB.
+
```
Smitsel: There is multiple descriptions about this field in the v2 endpoints. Let's pick one and use that consistently :)

This is the other one i found:
"Provide any data you like, and we will save the data alongside the customer. Whenever you fetch the customer with our API, we'll also include the metadata. You can use up to 1kB of JSON."

source/reference/v2/subscriptions-api/update-subscription.rst
```diff
@@ -79,6 +79,15 @@ example: ``/v2/customers/cst_5a2pPrwaWy/subscriptions/sub_8EjeBVgtEn``.
...
 
+   * - ``metadata``
+
+       .. type:: mixed
+          :required: false
+
+     - Provide any data you like, for example a string or a JSON object. We will save the data alongside the
+       subscription. Whenever you fetch the subscription with our API, we'll also include the metadata. You can use up
+       to approximately 1kB.
+
```
Smitsel: Same comment as above, lets pick one :)

source/reference/v2/subscriptions-api/create-subscription.rst
MollieRick: Done!

PR #235, opened 2018-09-19
PROC-1386 Documentation on discounts

willemstuursma:


willemstuursma:
@ndijkstra can you share this with the various module maintainers?

@Marvin-Magmodules @davdebcom fyi

source/orders/handling-discounts.rst
```diff
@@ -0,0 +1,149 @@
+Handling discounts
+==================
+
+Using the :doc:`Orders API </orders/overview>` you can pass the full details of your customers' orders to Mollie.
+Of course in some cases you might want to give the customer a discount. This document describes how you can pass
```
mvdpanne: I feel like there should be a comma after `Of course` 🤔 What do you think?

source/orders/handling-discounts.rst
```diff
@@ -0,0 +1,149 @@
...
+various types of discounts and promotions.
+
+
+2-for-1 type promotions
+^^^^^^^^^^^^^^^^^^^^^^^
+You may want to offer 2-for-1 type promotions, or other promotions where the consumer gets a certain percentage off if
+they buy multiple of the same product. In such cases, your customer gets some part of the order line for free.
```
mvdpanne: `multiple units of the same product`?

source/orders/handling-discounts.rst
```diff
@@ -0,0 +1,149 @@
...
+==================
+
+Using the :doc:`Orders API </orders/overview>` you can pass the full details of your customers' orders to Mollie.
+Of course in some cases you might want to give the customer a discount. This document describes how you can pass
+various types of discounts and promotions.
+
+
+2-for-1 type promotions
+^^^^^^^^^^^^^^^^^^^^^^^
+You may want to offer 2-for-1 type promotions, or other promotions where the consumer gets a certain percentage off if
```
mvdpanne: `consumer` or `customer`?

source/orders/handling-discounts.rst
```diff
@@ -0,0 +1,149 @@
...
+they buy multiple of the same product. In such cases, your customer gets some part of the order line for free. 
+
+You can do this using the ``discountAmount`` property on the ``lines`` array.
+
+An example of an order line for a 2-for-1 type promotion:
```
mvdpanne: Some small modifications I would make:
```
Here's an example of an order line that has a 2-for-1 type promotion:
```

source/orders/handling-discounts.rst
```diff
@@ -0,0 +1,149 @@
...
+               "value": "-10.00"
+           },
+           "vatAmount": {
+               "currency": "EUR",
+               "value": "0.00"
+           }
+       }
+   ]
+
+As a gift card is simply a means of payment and are untaxed, this does not affect the VAT amount charged to your
```
mvdpanne: Gift card is singular, so `are untaxed` should be `is untaxed`.

source/orders/handling-discounts.rst
```diff
@@ -0,0 +1,149 @@
...
+
+Note how the VAT is only calculated over the amount actually charged to your customer.
+
+Gift cards
+^^^^^^^^^^
+
+If Mollie does not handle your :doc:`gift cards</guides/gift-cards>` for you, you can add an additional line of the type
+``giftcard`` instead if your customer wants to apply a gift card. The line must then have a negative amount.
+
+An example where a shopper exchanges a €10.00 gift card:
```
mvdpanne: I would again go for `Here's an`. To me it makes the text read better from paragraph to paragraph. What do you think?

source/orders/handling-discounts.rst
willemstuursma: Customer. I want to move away from "consumer".

willemstuursma:
> Will gift card and store credit lines have to be shipped by the merchant?

Yes it has to be shipped, because we cannot create negative captures. We'll have to make sure the merchant doesn't forget to ship the discount line and just silently ship it if forgotten? What do you think?

mvdpanne:
Another thing that I'm not exactly sure about is when you have separate discount lines, the merchant is able to cancel these discount lines. What happens then?

A way around that would be to auto-ship such lines and make them non-refundable. This would make it more in line with the `discountAmount` parameter, as you can't get rid of that after creating the order either (right?), though that probably is too 'aggressive'.

It kind of depends on how we deal with e.g. a 2-for-1 deal of which one unit gets canceled. Or wait, that was for Phase II wasn't it? Then maybe this guide should also mention you cannot execute partial actions on order lines with a `discountAmount`.

willemstuursma:
>  Then maybe this guide should also mention you cannot execute partial actions on order lines with a discountAmount.

We'll have this soon enough, and it is already mentioned at the API reference docs.

PR #236, opened 2018-09-20
PROC-1395 Remove warnings from Orders API documentation

mvdpanne:
Please note that it was not removed from the `cancelOrder` endpoint, because the warning was never added to that endpoint.


shirshir:
Oh, it has some conflicts.

PR #239, opened 2018-09-24
Remove temporary redirects

vernondegoede:
Remove temporary redirects. Will now be handled by AWS Lambda functions.


hvalle:
LGTM.

PR #241, opened 2018-09-25
PROC-1306 - Find a good place on docs.mollie.com for our Reseller API

MollieRick:


willemstuursma:
Downloads can be moved to the Overview. 

It looks good to me. @vernondegoede what do you think?

Obviously, it can use a review and some polishing up.

vernondegoede:
@MollieRick Nice work 💪

I noticed something (unrelated) to this PR though: switching versions doesn't seem to work when you're in the Mollie Connect / Reseller API **Reference** docs. If the page does not exist using a v2 URL parameter, we could link to https://docs.mollie.com/reference/v2 or https://docs.mollie.com/reference/v1 (we recently added redirects from these URLs to the first article).

MollieRick:
@vernondegoede All endpoints are now into the docs. Can you check all text please? Some of them are very old and maybe not in the way how we want to communicate.

willemstuursma:
@MollieRick I'll review the wording.

willemstuursma:
`:doc:`obligatory parameters </reseller-api/secret-keys>``: this should use a `:ref:` so that it can link to the right header.

willemstuursma:
You should also proofread it yourself @MollieRick.

PR #255, opened 2018-10-01
A number of improvements to the Orders API documentation

mvdpanne:
General improvements
- Removed the `refunded` state for orders and order lines.
- `amountCaptured` and `amountRefunded` should only be returned in the order response if they are > 0.
- Added the `shippableQuantity`, `refundableQuantity` and `cancelableQuantity` properties for order lines in the response examples.
- `productUrl` and `imageUrl` are links, not direct properties of the order line.
- Improved the order state changes guide.

Specific improvements
- `cancelOrderlines` API can only be used with `authorized` or `shipping` lines.
- `createOrderRefund` API response should describe the partial order line, not the whole order line.

### DIFF
--- a/source/orders/status-changes.rst
+++ b/source/orders/status-changes.rst
@@ -50,7 +50,7 @@ The following diagram shows how one order status leads to another:
     that does not support authorizations.
 
     * Mollie will call your webhook when the order reaches this state.
-    * Order lines can be in the state ``paid``.
+    * All order lines will also be in the ``paid`` state.
     * Can transition to: ``shipping`` and ``completed``.
 
 .. _order-status-pending:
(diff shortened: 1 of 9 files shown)


source/orders/status-changes.rst
```diff
@@ -173,7 +178,7 @@ The following diagram shows how one order line status leads to another:
 ``completed``
 ^^^^^^^^^^^^^
     When the order line is completely shipped, it will get this status. The order line will also get this status when it
-    is partially shipped and the rest of the line is ``canceled``.
+    is e.g. partially shipped or refunded and the rest of the line is ``canceled``.
```
mvdpanne: This might make it more confusing, what do you think? 🤔

source/reference/v2/orders-api/get-order.rst
```diff
@@ -654,7 +685,17 @@ Response
...
-                "createdAt": "2018-08-02T09:29:56+00:00"
+                "createdAt": "2018-08-02T09:29:56+00:00",
+                "_links": {
+                    "productUrl": {
+                        "href": "https://shop.lego.com/nl-NL/Porsche-911-GT3-RS-42056",
+                        "type": "text/html"
+                    },
+                    "imageUrl": {
+                        "href": "https://sh-s7-live-s.legocdn.com/is/image/LEGO/42056?$PDPDefault$",
+                        "type": "text/html"
```
ibudasov: Looks strange that image has "type": "text/html", can it be something more suitable?

source/orders/status-changes.rst
```diff
@@ -136,12 +140,13 @@ The following diagram shows how one order line status leads to another:
...
     * The order has status ``created`` or ``pending``.
     * Can transition to: ``paid``, ``authorized`` and ``canceled``.
+    * Only the whole order can be canceled at this point in time, not individual order lines.
 
 .. _orderline-status-paid:
 
 ``paid``
 ^^^^^^^^
-    The order line status will be set to this status when the order's payment is successfully completed with a payment
+    The order line will be set to this status when the order's payment is successfully completed with a payment
```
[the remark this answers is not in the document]
mvdpanne: I'm not sure what you mean. Only `paid` flow orders should get in state `paid`.

source/reference/v2/orders-api/cancel-order.rst
```diff
@@ -74,22 +74,13 @@ Response
...
-        },
-        "amountRefunded": {
-            "value": "0.00",
-            "currency": "EUR"
-        },
         "status": "canceled",
         "isCancelable": false,
         "metadata": null,
         "createdAt": "2018-08-02T09:29:56+00:00",
         "expiresAt": "2018-08-30T09:29:56+00:00",
```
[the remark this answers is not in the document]
mvdpanne: You're right, `canceledAt` should be present.

source/reference/v2/orders-api/get-order.rst
mvdpanne: Ah, interesting. I guess it is this way because when you set an image related link type, e.g. `image/png`, the user of our API will have to provide a `.png` image specifically. Since we don't want to be that strict, we keep it at `text/html`.

Here's the list of all link types: https://www.iana.org/assignments/media-types/media-types.xhtml

PR #256, opened 2018-10-01
Use WGAC compliant color for syntax highlighting

### LINKED ISSUE #149: Improve PHP syntax highlighting

vernondegoede:
Fixes #149. Improved based on [the feedback in my previous PR](https://github.com/mollie/api-documentation/pull/218#issuecomment-420959663).

----

@sandervanhooft What do you think?

image

### DIFF
--- a/source/theme/styles/components/_highlight.scss
+++ b/source/theme/styles/components/_highlight.scss
@@ -81,10 +81,7 @@
 }
 
 .rst-content .highlight-bash .highlight .s2,
-.rst-content .highlight-bash .highlight .se {
-  color: $brand-primary;
-}
-
+.rst-content .highlight-bash .highlight .se,
 .rst-content .highlight-php .highlight .cp,
 .rst-content .highlight-php .highlight .nv,
 .rst-content .highlight-php .highlight .k,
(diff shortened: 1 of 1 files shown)


mattijsbliek:
Why is everything in the PHP code blue?

sandervanhooft:
@vernondegoede It's definitely an improvement! 👍 👍 

Didn't know about WGAC, nice.

My IDE is using more colors for php (is this what your comment is about @mattijsbliek ?), but I think this color combination strikes a nice balance between readability and the Mollie design colors. The php examples are simple, as in do not contain class or function definitions.

So it seems really good to me. My eyes are happy.

vernondegoede:
@mattijsbliek Design chose to make all requests (or code examples blue) and responses yellow. However, since the primary colour is not that readable, I checked what the closest WGAC compliant code was and made our syntax highlighting use that one.

Team Design will probably revamp this page in the future and introduce proper syntax highlighting. I believe this is fine until then.

mattijsbliek:
I think both are pretty bad to be honest, but yours is better than what’s currently live so ship it 🤷‍♂️

We should add proper syntax highlighting soon though.

PR #258, opened 2018-10-02
Moving guides to Reseller part

MollieRick:


MollieRick:
@willemstuursma Well, you can. There is now just one menu for guide and endpoints. This makes that all Reseller resources are together and accessable :)

PR #280, opened 2018-10-12
Add extra category codes

BertvanHoekelen:


willemstuursma:
Please keep all list alphabetical - or in this case, ordering by MCC might be more appropriate.

PR #288, opened 2018-10-19
Add discount order line information and example

ndijkstra:


willemstuursma:
Isn't it sufficient to update the title of _Store credit and vouchers_ to _Adding a discount to the whole order_

PR #293, opened 2018-10-23
PROC-1453 Add link for changing recurring payment state

mvdpanne:
- `checkout` url is optional, because recurring payment do not have it.
- The `changePaymentState` url is used to change the payment state of recurring payments.

### DIFF
--- a/source/changelog/v1/changelog.rst
+++ b/source/changelog/v1/changelog.rst
@@ -9,7 +9,7 @@ October 2018
 Wednesday, 24th
 ---------------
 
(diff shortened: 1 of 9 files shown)


willemstuursma:
Can you add some details to https://docs.mollie.com/payments/recurring

willemstuursma:
Can you add it to the changelog?

willemstuursma:
I think we could use a dedicated page on testing the Mollie API. Now there are two paragraphs in the Authentication page that mention this. Can you move it around and create a dedicated page? And then some more info can be left in here.

PR #295, opened 2018-10-24
Jorik mollie v1 chb negative

JorikMollie:


willemstuursma:
I guess that by now, it is a [Common Law Feature](https://blog.codinghorror.com/new-programming-jargon/).

PR #297, opened 2018-10-25
Update cancel status changes on order

willemstuursma:
Follows https://github.com/mollie/magento2/issues/83.

### DIFF
--- a/source/orders/migrate-to-orders.rst
+++ b/source/orders/migrate-to-orders.rst
@@ -99,11 +99,11 @@ Just like in the payments API you can specify a ``webhookUrl`` that will be used
 when the status of an order has changed. You can then use the Mollie API to
 :doc:`retrieve the order status </reference/v2/orders-api/get-order>`.
 
-Note that orders cannot be canceled by shoppers. The order will remain ``created`` so that you can add further payments
-to the order to give your customer a second chance to pay for the order.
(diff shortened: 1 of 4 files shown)


source/orders/migrate-to-orders.rst
```diff
@@ -93,21 +93,28 @@ orders' statuses.
...
 
-Retrieving payment methods
---------------------------
+Receiving status updates
+------------------------
+Just like in the payments API you can specify a ``webhookUrl`` that will be used by Mollie to inform your back office
+when the status of an order has changed. You can then use the Mollie API to
+:doc:`retrieve the order status </reference/v2/orders-api/get-order>`.
+
+Note that orders cannot be canceled by shoppers. The order will remain ``created`` so that you can add further payments
```
mvdpanne: I think the following:
```
... so that you can create additional payments for the order ...
```
sounds better than:
```
... so that you can add further payments to the order ...
```

source/orders/migrate-to-orders.rst
```diff
@@ -93,21 +93,31 @@ orders' statuses.
...
+to the order to give your customer a second chance to pay for the order.
+
+If you want to know if your customer canceled the first payment, you will need to retrieve the payment together with
+the order instead of just the order by adding a ``?embed=payments`` to the Get Order API request. You can then find the
```
mvdpanne: `adding a` --> the `a` can be dropped.

PR #309, opened 2018-11-06
Implement mollie.com navigation on docs.mollie.com

vernondegoede:
Implemented new design from mollie.com for docs.mollie.com.


vernondegoede:
Fonts are currently not loading locally, need to change some nginx settings for our CDN first.

PR #312, opened 2018-11-08
PROC-1587 - Set lifetime test subscriptions to canceled after 10 times

MollieRick:


source/reference/v1/subscriptions-api/create-subscription.rst
```diff
@@ -57,6 +57,10 @@ Replace ``customerId`` in the endpoint URL by the customer's ID, for example
 
      - Total number of charges for the subscription to complete. Leave empty for an ongoing subscription.
 
+       .. note::
+          Subscriptions in test mode will be canceled automatically after 10 times when you do not specify a number of
```
willemstuursma: @MollieRick what if you specify 11 charges? Let's max out at 10

MollieRick: @willemstuursma It will only cancel after 10 times when there is no `times` set. Otherwise it will charge up till the given `times`.

PR #320, opened 2018-11-13
Add language switch to request examples

albert-carreras:
### Adds a convenient language switcher to the request examples

- Adds new code block directive `code_block_selector_directive.py` that uses `.. code-block-selector::`  to detect the need for the selector and adds a few html elements:
   - A switcher container
   - An empty div (with `nodes.raw`) which only purpose is to have the `data-enhancer` attribute visible
   - The Switcher buttons
   - The code-blocks with new ids to find them on the DOM

- Adds a new data-enhancer handler that deals with the hiding and showing of codeblocks, remembering selected language among others.
- Adds the new directive to all relevant `.rst` files
- Improves responsiveness in all pages


source/reference/v2/captures-api/get-capture.rst
```diff
@@ -152,25 +152,22 @@ Response
...
-^^^^^^^^^^^^^
-.. code-block:: php
-   :linenos:
+   .. code-block:: php
+    :linenos:
 
-    <?php
-    $mollie = new \Mollie\Api\MollieApiClient();
-    $mollie->setApiKey("test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM");
+        <?php
```
mvdpanne: This indentation is not necessary/adds space to the code example.

source/reference/v2/captures-api/list-captures.rst
```diff
@@ -103,25 +103,23 @@ Response
...
+    curl -X GET https://api.mollie.com/v2/payments/tr_WDqYK6vllg/captures \
+        -H "Authorization: Bearer test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM"
+
+   .. code-block:: php
+    :linenos:
 
-    <?php
-    $mollie = new \Mollie\Api\MollieApiClient();
-    $mollie->setApiKey("test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM");
+        <?php
```
mvdpanne: Please check all PHP examples for this additional indentation.

source/reference/v2/captures-api/get-capture.rst
```diff
@@ -152,25 +152,22 @@ Response
 Example
 -------
 
-Request (curl)
-^^^^^^^^^^^^^^
-.. code-block:: bash
-   :linenos:
+.. code-block-selector::
+   .. code-block:: bash
+    :linenos:
```
mvdpanne: The `:linenos:` does not seem to be aligned correctly. The `Response` examples on show the correct indentation of the directive arguments and body. You can also see the example here: https://mollie.atlassian.net/browse/PROC-1601

It's kinda a pain to edit it all, but RST is quite sensitive to whitespace, so I'd say it's worth it to keep it strict and clean.

source/theme/js/exampleSwitcher.js
```diff
@@ -0,0 +1,38 @@
+import { enhance } from './utils';
+
+// Gets all the codeblocks present and hides all of them except the one from the argument
+const hideCodeBlocks = language => {
+  const codeBlocks = Array.from(document.querySelectorAll('[id^="request-"]'));
+  codeBlocks.forEach(block => {
+    const elementId = `request-${language.toLowerCase()}`;
+    if (block.id !== elementId) block.classList.add('hidden');
```
mvdpanne: Is it default code style in JS not to use curly braces for conditional statements?

source/theme/js/exampleSwitcher.js
```diff
@@ -0,0 +1,38 @@
...
+    else block.classList.remove('hidden');
+  });
+};
+
+export default enhance('example-switcher', element => {
+  // Gets stored language if present
+  const storedLanguage = localStorage.getItem('preferedLanguage');
+  const language = storedLanguage && storedLanguage !== 'undefined' ? storedLanguage : 'curl';
```
mvdpanne: This feels a little crammed together, but maybe that's just me.

source/theme/js/exampleSwitcher.js
vernondegoede: Nope, we always use curly brackets at Mollie. Can you make sure this uses curly brackets as well?

source/theme/js/exampleSwitcher.js
```diff
@@ -0,0 +1,46 @@
...
+      block.classList.add('hidden');
+    } else {
+      block.classList.remove('hidden');
+    }
+  });
+};
+
+export default enhance('example-switcher', element => {
+  // Gets stored language if present
+  const storedLanguage = localStorage.getItem('preferedLanguage');
```
vernondegoede: It should be `preferredLanguage` 🥇

source/theme/js/exampleSwitcher.js
```diff
@@ -0,0 +1,46 @@
...
+  let language;
+  if (storedLanguage && storedLanguage !== 'undefined') {
+    language = storedLanguage;
+  } else {
+    language = 'curl';
+  }
+  const selectedLanguage = document.getElementById(`example-switch-${language}`);
```
vernondegoede: Can you add a check here and see if the element is available? If not, fall back to `DEFAULT_LANGUAGE` (which you can re-use in the `} else {` on line 23.

vernondegoede: The reason we need to check this, is because I might have selected "Ruby" and navigate to a page where there is no Ruby example.

source/theme/styles/components/_example.scss
```diff
@@ -1,3 +1,49 @@
...
+      &.selected {
+        font-weight: 600;
+        border-bottom: none;
+      }
+    }
+  }
+
+  .code-blocks > div {
+
+    &:not(.hidden) {
```
vernondegoede: Can you use `.is-hidden` here? Hidden is a stateful classname, so it's allowed to break the BEM rules.

source/theme/styles/components/_example.scss
```diff
@@ -1,3 +1,49 @@
...
+      cursor: pointer;
+
+      &:hover,
+      &:focus {
+        color: $brand-primary-hover;
+        text-decoration: none;
+        border-bottom: none;
+      }
+
+      &.selected {
```
vernondegoede: `.is-selected`

source/theme/styles/components/_example.scss
```diff
@@ -1,3 +1,49 @@
+.rst-content #example {
+  max-width: 700px;
+
+  .examples-switcher {
+
```
vernondegoede: Nitpick: Can you remove the whitespace? :)

source/theme/styles/components/_example.scss
```diff
@@ -1,3 +1,49 @@
...
+      &.selected {
+        font-weight: 600;
+        border-bottom: none;
+      }
+    }
+  }
+
+  .code-blocks > div {
+
```
vernondegoede: Nitpick: Unnecessary whitespace

source/theme/js/exampleSwitcher.js
```diff
@@ -0,0 +1,46 @@
+import { enhance } from './utils';
+
+// Gets all the codeblocks present and hides all of them except the one from the argument
+const hideCodeBlocks = language => {
+  const codeBlocks = Array.from(document.querySelectorAll('[id^="request-"]'));
+  codeBlocks.forEach(block => {
+    const elementId = `request-${language.toLowerCase()}`;
+    if (block.id !== elementId) {
+      block.classList.add('hidden');
```
vernondegoede: Can you also add an `aria-hidden` attribute here?

source/theme/styles/main.scss
```diff
@@ -188,10 +188,10 @@ a.headerlink::before {
 @import 'components/code';
 @import 'components/content';
 @import 'components/endpoint';
+@import 'components/example';
```
vernondegoede: Could you rename this to `code-example`?

vernondegoede:
`localStorage` is not available in Safari private mode if I remember correctly. Can we add a check for this and always just show the first item (instead of `getItem`) if it isn't available?

More information: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Testing_for_availability

source/orders/overview.rst
```diff
@@ -14,7 +14,7 @@ How does the Orders API work?
 #. For every order in your webshop, you create an order using the
    :doc:`Create Order API </reference/v2/orders-api/create-order>`.
 
-#. The :ref:`Create Order API response <get-order-response>` contains the ``_links.checkout`` property. This is a link
+#. The :ref:`Create Order API response <create-order-response>` contains the ``_links.checkout`` property. This is a link
```
vernondegoede: Sharp 💯

source/theme/js/exampleSwitcher.js
albert-carreras: 👀

mvdpanne:
Ah, one last thing I just noticed: the width of the code examples does always not match up with the width of the parameter table. When you widen your browser, the code example does not get as wide as the parameter table.

PR #324, opened 2018-11-16
Add "klarnasliceit" and "klarnapaylater" to possible method values in GET payment

vernondegoede:
It is possible to retrieve a payment created by an order that uses Klarna.


willemstuursma:
Please keep the list sorted a-z

vernondegoede:
Done! ✅

PR #325, opened 2018-11-16
PROC-1632 - Adding guides for Mobile usage

MollieRick:


source/mobile-apps/getting-started-connect.rst
```diff
@@ -0,0 +1,186 @@
...
+Step 3: Create a server-side file for Access-tokens
+---------------------------------------------------
+For safety-reasons you don't want the Client Secret inside your app. Therefor you should create a file on your server
+that converts the authentication code (what you get from the OAuth authorize-screen) to an actual Access-token what you
+can call the Mollie API with. Make sure the line between your app and your server is **secure and not accessable for someone
+else except your app**. Please see the :doc:`OAuth documentation </reference/oauth2/tokens>` how to set this up.
+
+Step 4: Let your App open the authorization page
+------------------------------------------------
+Let's assume that you put a login button in your app that needs to open the Mollie OAuth flow. Add the following code to
```
willemstuursma: Should have a note or indication that you need to randomly generate the state.

willemstuursma:
Some generic guide lines:

* You'll (and other abbreviations) are for speaking language, written language should always write this in full.

willemstuursma:
Some remarks:

* The app should not pass the amount to the backend service. An impersonator could send € 0,01 and get stuff on the cheap.
* The server should set up a webhook where Mollie sends status updates. When the payment is completed, the server should send the app a silent push notification upon which the app reacts again (e.g. by showing an order confirmation screen). Can you add some directions on this.

MollieRick:
@willemstuursma 
> I think the Kotlin examples make it a bit noisy. They are quite similar to the Java examples.
True, but from many sides you will get the advise to start a new project with Kotlin, instead of Java. What do you suppose? Removing the Kotlin snippets?

willemstuursma:
> True, but from many sides you will get the advise to start a new project with Kotlin, instead of Java. What do you suppose? Removing the Kotlin snippets?

It's more about the general flow than the exact code.

MollieRick:
How about the current guides @willemstuursma ?

willemstuursma:
Thanks @MollieRick. What we have now, is a good start for app developers. We'll see, what happens next and what we will need to improve.

PR #330, opened 2018-11-20
PROC-1488 Update documentation regarding recurring orders

willemstuursma:


source/reference/v2/payments-api/create-payment.rst
```diff
@@ -77,7 +77,7 @@ Parameters
        the order when your customer returns.
 
        .. note::
-          For payments with ``sequenceType`` ``recurring``, you can skip this parameter. For all other payments, this
+          For payments with ``sequenceType`` ``recurring``, you can omit this parameter. For all other payments, this
```
shirshir: I guess the `required: true` should now be changed as well?

willemstuursma: I updated the documentation.

PR #331, opened 2018-11-20
Chargebacks: pagination info, improved php example

sandervanhooft:
Added example for listing all chargebacks using the (root) chargebacks endpoint.

The new endpoint is not yet released.

### DIFF
--- a/source/reference/v2/chargebacks-api/list-chargebacks.rst
+++ b/source/reference/v2/chargebacks-api/list-chargebacks.rst
@@ -125,6 +125,7 @@ Example
       $chargebacks = $payment->chargebacks();
 
       // List chargebacks across all payments on the payment profile
+      // (For all chargebacks on the organizations, use an OAuth or Organization access token.)
       $all_chargebacks = $mollie->chargebacks->page();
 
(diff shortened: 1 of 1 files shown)


source/reference/v2/chargebacks-api/list-chargebacks.rst
```diff
@@ -102,9 +120,13 @@ Example
       $mollie = new \Mollie\Api\MollieApiClient();
       $mollie->setApiKey("test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM");
 
+      // Listing chargebacks for a single payment
       $payment = $mollie->payments->get("tr_7UhSN1zuXS");
       $chargebacks = $payment->chargebacks();
 
+      // Listing chargebacks across all payments
```
willemstuursma: ...all payments _on the payment profile_. 

For all chargebacks on the organizations, use an OAuth or Organization access token.

sandervanhooft: Adding "For all chargebacks on the organizations, use an OAuth or Organization access token." in next commit

PR #333, opened 2018-11-26
Friendlier contribute msg

dafeso:
Friendlier contribute msg


source/extensions/mollie/code_block_selector_directive.py
```diff
@@ -100,8 +100,8 @@ def extract_code_examples(self, code_blocks):
     def create_generic_code_block(self, lexer):
         properties = self.supported_lexers[lexer]
         content = (
-            "We do not yet have a {} code example for this API call.\n" +
-            "If you have some spare time, you can open a pull request at:\n" +
+            "We don't have a {} code example for this API call yet.\n\n" +
+            "If you have some time to spare, feel free to open pull request at:\n" +
```
Smitsel: This should be "open a pull request".

I like the friendlier wording though!

PR #339, opened 2018-12-03
Mollie *does* call the webhook for the "created" order state.

vernondegoede:
Update documentation to reflect that the webhook will be called to the 'created' order status change.


willemstuursma:
This shouldn't happen, there is nothing to report on.

PR #342, opened 2018-12-06
PROC-1670 - Possibility to add webhookUrl to order payments

MollieRick:


willemstuursma:
It can use some more information on who / why it should be used.

PR #345, opened 2018-12-10
Adds Payment method switcher

albert-carreras:
Needs to be created with JS and appended to the DOM as we want it in the doc itself (that is compiled from the `.rst` files):

1) We create a dropdown element with the data-enhancer attribute.
   - This dropdown is generated based on the DOM, as we want to select based on payment methods present in that example
   - We use the existing content, such as names and ids for the methods to fill the options of the select.
   - We place the dropdown just before the first div in the payments method block, after the block's title and description. We have the method's name repeated but this compromise was the best solution.
2) We call the function that creates that dropdown before the enhancers runs.
3) The dropdown enhancer is called, just like the rest.
   - This enhancer hides and shows payment method blocks according to the select.
   - This enhancer saves the selected method to the local storage.

### DIFF
--- a/source/theme/js/linkDropdownPaymentMethods.js
+++ b/source/theme/js/linkDropdownPaymentMethods.js
@@ -1,0 +1,53 @@
+import { enhance } from './utils';
+import Dropkick from 'dropkickjs';
+
+const DEFAULT_METHOD = 'bancontact';
+
+export const linkDropdownPaymentMethods = enhance('payment-method-switcher', element => {
+  element.addEventListener('change', () => {
+    element.childNodes.forEach(method => {
+      const selectedElement = document.getElementById(method.value);
+      if (method.value !== element.value) {
+        selectedElement.classList.add('u-screenreader-only');
+        selectedElement.setAttribute('aria-hidden', 'true');
+      } else {
+        selectedElement.classList.remove('u-screenreader-only');
+        selectedElement.setAttribute('aria-hidden', 'false');
+        localStorage.setItem('preferredPaymentMethod', method.value);
+      }
+    });
+  });
+
+  new Dropkick(element, {
+    mobile: false,
+  });
+});
+export const addPaymentMethodDropdown = () => {
+  const paymentMethodsBlock = document.querySelectorAll('[id^=payment-method]');
+  const storedMethod = localStorage.getItem('preferredPaymentMethod');
+  let preferredMethod;
+  if (storedMethod && storedMethod !== 'undefined') {
+    preferredMethod = storedMethod;
+  } else {
+    preferredMethod = DEFAULT_METHOD;
+  }
+  if (paymentMethodsBlock.length) {
+    const paymentMethods = paymentMethodsBlock[0].childNodes;
+    let options = '';
+    paymentMethods.forEach(method => {
+      if (method.nodeName === 'DIV') {
+        const isSelected = method.id === preferredMethod;
+        if (!isSelected) {
+          method.classList.add('u-screenreader-only');
+          method.setAttribute('aria-hidden', 'true');
+        }
+        const methodName = method.childNodes[1].textContent.replace('¶', '');
+        options += `<option value="${method.id}" ${
+          isSelected ? ' selected' : ''
+        }>${methodName}</option>`;
+      }
+    });
+    const dropdown = `<select data-enhancer="payment-method-switcher" class="link-dropdown">${options}</select>`;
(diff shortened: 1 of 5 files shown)


mattijsbliek:
z-index conflict with the sub navigation:

image

source/theme/js/linkDropdownPaymentMethods.js
```diff
@@ -0,0 +1,85 @@
...
+const DEFAULT_METHOD = 'bank-transfer';
+
+// Adds the onchange listener that hides/shows methods based on selection
+export const linkDropdownPaymentMethods = enhance('payment-method-switcher', element => {
+  element.addEventListener('change', () => {
+    element.childNodes.forEach(method => {
+      const selectedElement = document.getElementById(method.value);
+      if (method.value !== element.value) {
+        selectedElement.classList.add('u-screenreader-only');
+        selectedElement.setAttribute('aria-hidden', 'true');
```
mattijsbliek: Ok so this is an interesting problem, the above two attributes conflict. `.u-screenreader-only` allows content to remain visible for screenreaders (and Google), but `aria-hidden="true"` hides it.

On top of this, links in hidden content are still able to receive focus, so tabbing through the page is a weird experience. We could remove `.u-screenreader-only` and use `hidden="true"` to hide the content, but that would also hide it from Google. I think what we probably want to do here is use the `inert` attribute with a [polyfill](https://github.com/wicg/inert).

And maybe `aria-expanded="false"` would be preferred over `aria-hidden`? I think this a tricky one to make properly accessible while keeping content visible for Google.

source/theme/js/linkDropdownPaymentMethods.js
```diff
@@ -0,0 +1,85 @@
...
+  });
+});
+
+// Adds the payment method dropdown to the DOM
+// We need to do this with JS, as the doc file is compiled from the .rst files
+export const addPaymentMethodDropdown = () => {
+  //Gets all possible payment methods and the preferred one from storage
+
+  // We need to get the payment methods block (different names in different pages, same prefix)
+  const paymentMethodsBlockArray = document.querySelectorAll('[id^=payment-method]');
```
mattijsbliek: This is an awfully broad selector. We might want to try and scope it a bit to prevent it from finding unwanted elements. Also, it’s a NodeList not and array, might be misleading to other devs. Would rename to simply `paymentMethods` :)

source/theme/js/linkDropdownPaymentMethods.js
```diff
@@ -0,0 +1,85 @@
...
+
+  if (paymentMethodsBlockArray.length) {
+    const paymentMethodsBlock = paymentMethodsBlockArray[0];
+
+    // Get the prefered method, either from storage or the default one.
+    const storedMethod = localStorage.getItem('preferredPaymentMethod');
```
mattijsbliek: I believe localStorage is not available when browsing on Safari in private mode. We could extract everything related to storing a preferredMethod to a separate function, so we can early return when we don’t have access to localStorage.

source/theme/js/linkDropdownPaymentMethods.js
```diff
@@ -0,0 +1,85 @@
...
+      preferredMethod = storedMethod;
+    } else {
+      preferredMethod = DEFAULT_METHOD;
+    }
+
+    let options = '';
+    let firstMethod;
+
+    // Adds the selector option for each method
+    paymentMethodsBlock.childNodes.forEach(method => {
```
mattijsbliek: IE11 doesn’t support forEach on a NodeList and I’m pretty sure we don’t have a polyfill for it. I would be fine with early returning when NodeList.prototype.forEach is undefined and not enhancing with a dropdown.

source/theme/js/linkDropdownPaymentMethods.js
```diff
@@ -0,0 +1,85 @@
...
+      // Inside the block, we are interested in the divs, not the title nor the description
+      if (method.nodeName === 'DIV') {
```
mattijsbliek: Can you flip this and early return so we don’t have nested if statements?

mattijsbliek:
Would be good to check `window.location.hash` to see if somebody is trying to access a certain method, and show that one if it’s the case.

source/theme/js/linkDropdownPaymentMethods.js
mattijsbliek: Ah that’s good 👍 They used to throw an error which was super annoying.

PR #352, opened 2018-12-13
Fixing wrong links

MollieRick:


Makefile
```diff
@@ -50,15 +50,15 @@ lint-js:
 # This checks for links that are missing the trailing underscore. They are valid reStructured text but probably not your
 # intention.
 verify:
-	! find source -name '*.rst' | xargs grep --color -E '<http.*>`([^_]|$$)'
+	! find source -name '*.rst' | xargs grep --color -E '<http.*>`([^_]|$$)' | grep 'mollie\.test' -r  source/ | grep 'mollie\.dev' -r  source/
```
willemstuursma: I think piping it around is not the right way
Better to move it to dedicated lines

source/reference/v2/profiles-api/enable-method.rst
```diff
@@ -59,13 +59,17 @@ Response
        "id": "bancontact",
        "description": "Bancontact",
        "image": {
-           "size1x": "https://mollie.dev/external/icons/payment-methods/bancontact.png",
-           "size2x": "https://mollie.dev/external/icons/payment-methods/bancontact%402x.png",
-           "svg": "https://mollie.dev/external/icons/payment-methods/bancontact.svg"
+           "size1x": "https://mollie.com/external/icons/payment-methods/bancontact.png",
```
willemstuursma: www is missing

PR #358, opened 2018-12-18
Payment switcher fix for non existing anchors

albert-carreras:
Navigating to this page with a non payment method anchor resulted in not showing a proper payment method.

This checks that the method exists before using it.

### DIFF
--- a/source/theme/js/linkDropdownPaymentMethods.js
+++ b/source/theme/js/linkDropdownPaymentMethods.js
@@ -42,7 +42,7 @@ export const addPaymentMethodDropdown = () => {
 
     let preferredMethod;
 
-    if (hashMethod && !!paymentMethodsBlockArray.some(m => m.id === hashMethod)) {
+    if (hashMethod && paymentMethodsBlockArray.some(m => m.id === hashMethod)) {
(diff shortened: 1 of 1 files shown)


source/theme/js/linkDropdownPaymentMethods.js
```diff
@@ -42,7 +42,7 @@ export const addPaymentMethodDropdown = () => {
 
     let preferredMethod;
 
-    if (hashMethod) {
+    if (hashMethod && paymentMethodsBlockArray.some(m => m.id === hashMethod)) {
```
mattijsbliek: Technically I think you want to use `array.find` here, but maybe that’s nitpicking 🤷‍♂️

albert-carreras: some is more efficient, as it stops as soon as it finds it, it also returns a boolean which is more convenient

mattijsbliek: The efficiency is the same, Array.find also early returns as described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Description 🙂

PR #375, opened 2019-01-10
APM-52 - Add an API endpoint for retrieving all API Payment Methods

MollieRick:


source/reference/v2/methods-api/list-all-methods.rst
```diff
@@ -0,0 +1,81 @@
...
+If you are using :doc:`organization access tokens </guides/authentication>` or are creating an
+:doc:`OAuth app </oauth/overview>`, the following query string parameters are also available. With the ``profileId``
+parameter, you must specify which profile you want to look at when listing payment methods. Organizations can have
+multiple profiles for each of their websites. See :doc:`Profiles API </reference/v2/profiles-api/get-profile>` for more
+information.
+
+.. list-table::
+   :widths: auto
+
+   * - ``profileId``
```
willemstuursma: These parameters don't make sense to me.

MollieRick: It is necessary for the `issuers`.

Smitsel: I've discussed this with Rick f2f, it can indeed be omitted and purely done on organization level.

PR #379, opened 2019-01-12
Description for the `_embedded` prop on the Get Order page

firstred:
This PR adds a description for the `_embedded` property in the Get Order response.


Smitsel:
Thanks!

PR #383, opened 2019-01-14
APM-88 Metadata allthethings

Smitsel:


source/reference/v2/orders-api/create-order.rst
```diff
@@ -791,6 +800,7 @@ Response
                "sku": "5702016116977",
                "type": "physical",
                "status": "created",
+               "metadata": null,
```
MollieRick: Please add at least one example of how we return metadata.

PR #384, opened 2019-01-15
Get Order: `amountCaptured` is optional in response

firstred:
An order with status `created` does not have the `amountCaptured` field.


Smitsel:
Thanks!

PR #392, opened 2019-01-21
Add maximum interval

JorikMollie:
The API docs were missing information about the maximum interval for subscriptions.


willemstuursma:
Guys are you even reading these PR's. Please change the number of days in a year to the correct number. 😔

PR #400, opened 2019-01-25
APM-141 - Removing and changing references to mandateReference

MollieRick:


source/reference/v1/mandates-api/get-mandate.rst
```diff
@@ -162,7 +162,7 @@ Response
 
        .. type:: string
 
-     - The mandate's custom reference, if this was provided when creating the mandate.
+     - The mandate's custom reference. This parameter is omitted if the mandate does not have a custom reference.
```
Smitsel: I'm not sure if this is the right description. We've omitted it for all mandates from xx-xx-xxxx.
But maybe @ndijkstra has some insights for it?

MollieRick:
Not longer the case

PR #404, opened 2019-01-30
Remove `Set Fees` endpoint from the sidebar navigation.

mattijsbliek:
According to our sales department, this endpoint is only accessible for a handful of partners. They would like to hide it from the navigation to avoid confusing regular resellers who think this functionality will be available to them.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -65,6 +65,9 @@ todo_include_todos = False
 
 # -- Options for HTML output ----------------------------------------------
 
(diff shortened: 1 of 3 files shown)


lvgunst:
Can we also disable search engine indexing for this page?

mattijsbliek:
Good idea, added this. You can now add pathnames to the `do_not_index` list in the global configuration file.

PR #418, opened 2019-02-14
Add note about quantity and discountAmount

davdebcom:
@ndijkstra Do you agree?


willemstuursma:
I don't think this is right, we made some changes to allow this @davdebcom.

What problem were you running into?

davdebcom:
@willemstuursma The issue was eventually unrelated, but Nick told me it was not allowed so I thought it might be good to add to the docs. If it's allowed now, I'll close this PR.

Issue #410, opened 2019-02-04
Add code samples for each API client to Pagination guide


vernondegoede:
We would like to show code samples showing how the pagination works for our various API clients to [the Pagination guides](https://github.com/mollie/api-documentation/blob/master/source/guides/pagination.rst).

To do:
- Add example on how to navigate between pages using API client for:
  - [ ]   Node.js (@firstred)
  - [x]   PHP (@sandervanhooft)
  - [ ]   Python
  - [ ]   Ruby (@justincase)
- Remove pagination examples from code samples in API reference.

sandervanhooft:
@willemstuursma Will do!

sandervanhooft:
I've just PR'd the php v2 example, but it seems to be breaking the code block selector.

May be related to having multiple examples on one page with the same html ID tag (had that one before, not sure at what page/what the fix was).

Any help?

PR #421, opened 2019-02-21
Pagination php example

sandervanhooft:
#410

Code block selector fixed in #426 by @albert-carreras .

### DIFF
--- a/source/guides/pagination.rst
+++ b/source/guides/pagination.rst
@@ -70,11 +70,33 @@ Example of v2 pagination
 
 Request
 """""""
-.. code-block:: bash
-   :linenos:
+.. code-block-selector::
+   .. code-block:: bash
+      :linenos:
 
-   curl -X GET https://api.mollie.com/v2/payments \
-       -H "Authorization: Bearer test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM"
+      curl -X GET https://api.mollie.com/v2/payments \
+          -H "Authorization: Bearer test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM"
+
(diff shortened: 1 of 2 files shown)


sandervanhooft:
@albert-carreras 

This issue seems familiar to the one that you fixed here (enable profile method):

- https://github.com/mollie/api-documentation/commit/cea27849f3ec428a9a0d90270359d293b46c760c
- https://github.com/mollie/api-documentation/commit/43883d6bc0b1eb639b1934cef40c44de28341cfd

I have no idea how this build is being processed or how the code block selector works, but it seems `link`/`id` related.

Perhaps we could circumvent the issue by adding a prefix like @mvdpanne does [here](https://github.com/mollie/api-documentation/pull/422#issue-255316032):

```
* - .. param-name:: billingEmail
       :prefix: bankTransfer
```

source/guides/pagination.rst
```diff
@@ -70,11 +70,33 @@ Example of v2 pagination
```
albert-carreras: For the code-block-selector to work, the title needs to be consistent with the rest of examples.
Like this:

```
Example
-------

.. code-block-selector::
```

sandervanhooft: I understand, but what if we want to have a more distinguishable title? Is this possible?

V1 vs V2 is an important difference, should not be overlooked.

albert-carreras: Hmm I agree!

https://github.com/mollie/api-documentation/pull/426

vernondegoede:
@sandervanhooft I believe this one can be merged.

sandervanhooft:
@vernondegoede Reverted `package-lock.json`.

Issue #420, opened 2019-02-21
Add anchors on parameter level


sandervanhooft:
For the Cashier documentation I'm pointing to Mollie's API reference for some specific details.

In the case of "Read more about locale configuration options here", I'd like to point the developers directly to the `locale` parameter, instead of the Parameters heading (this page https://docs.mollie.com/reference/v2/payments-api/create-payment#parameters), since these parameter lists can be extensive.

Currently there are no anchors on this level.

**_So jump directly here:_**

<img width="604" alt="image" src="https://user-images.githubusercontent.com/7265703/53176648-d09dc580-35ee-11e9-9802-46e56c55e074.png">

**_Instead of here:_**

<img width="604" alt="image" src="https://user-images.githubusercontent.com/7265703/53176697-e8754980-35ee-11e9-84bd-cda316fd4dc9.png">

Is this possible? Would be really convenient!

willemstuursma:
@vernondegoede ? any ideas?

sandervanhooft:
The anchor could be something like `#parameters-locale`

vernondegoede:
I agree, this is annoying. Ran into this a couple of times myself. We should add a custom Sphinx directive which adds an `id` to a `div`, which would fix this.

@mvdpanne Any idea how much work this would be?

mvdpanne:
Put something together, seems to work.

I'll add a PR tomorrow.

vernondegoede:
@sandervanhooft You can now link them to https://docs.mollie.com/reference/v2/payments-api/create-payment#parameter-locale

sandervanhooft:
@mvdpanne @vernondegoede Awesome, thanks! 😎

PR #422, opened 2019-02-22
Add parameter anchor links

mvdpanne:
As requested in issue #420, this PR adds parameter-level anchor links. This is done by adding a new custom directive for the parameter name. It adds an id to the container in which the parameter name resides, and allows for specifying a prefix, such that parameters with identical names can still be referred to uniquely.

To put this into practice, instead of this:
```
* - ``amount``
```
We can now use this:
```
* - .. param-name:: amount
```
Which results in the `#parameter-amount` anchor tag to link to the amount parameter.

As mentioned, you can add a prefix to uniquely identify parameters that occur multiple times on the same page, such as `billingEmail`, like so:
```
* - .. param-name:: billingEmail
       :prefix: bankTransfer
```
Which results in the `#parameter-bankTransfer-billingEmail` anchor tag to link to the `billingEmail` parameter under the bank transfer section.

**Please note**:
- For now, I've only added the new syntax to three parameters. I'd like to get feedback on this, and if it's good to go, we can do a bulk search and replace to apply it everywhere. Prefixes will have to be added manually, I think.
- We will probably need to make some modifications on the front end to allow end users to easily copy this anchor link, just like for the headers on each page.

### DIFF
--- a/source/extensions/mollie/param_name_directive.py
+++ b/source/extensions/mollie/param_name_directive.py
@@ -1,0 +1,34 @@
+from docutils import nodes
+from docutils.parsers.rst import Directive, directives
+
+
+class ParamNameDirective(Directive):
+    has_content = False
+    optional_arguments = 1
+    final_argument_whitespace = True
+    option_spec = {
+        "prefix": directives.unchanged_required
+    }
+
+    def run(self):
+        param_name = self.arguments[0]
+        param_anchor_text = self.get_param_anchor_text(param_name)
+
+        container = nodes.container()
+        container["ids"].append(param_anchor_text)
+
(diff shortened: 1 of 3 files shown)


vernondegoede:
Super nice, thanks!

mvdpanne:
I suppose we'd want to do a bulk search and replace for this at some point?

Or do we want to wait for more demand before applying it everywhere? I'm not sure if @sandervanhooft meant to use it in one place only.

sandervanhooft:
A bulk search/replace would be best IMHO. 

I think it should be site-wide, but right now I only need a few parameter links. I can add them manually, but doing so (incl PRs etc) would make it very time consuming, especially in the long run.

PR #425, opened 2019-02-28
ONB-318 profile website validation

Smitsel:


source/reference/reseller-api/endpoints/profile-create.rst
```diff
@@ -52,7 +52,9 @@ parameters:
        .. type:: URL
           :required: true
 
-     - The url of the website profile
+     - The url of the website profile. The URL must be compliant to `RFC3986 <https://tools.ietf.org/html/rfc3986>`_
```
lvgunst: We also do not support URLs with a `@`. Should we add this?

Smitsel: Yes we should, thanks.

PR #428, opened 2019-03-08
Add missing failure reasons

JorikMollie:
Two credit card failure reasons were missing in the docs. Added them.


mvdpanne:
I already approved it, but I just saw this:
```
```possible_fraud```
```
Which should be:
```
``possible_fraud``
```

PR #432, opened 2019-03-14
Add new payment method Przelewy24 to the changelog.

shirshir:


Smitsel:
Guess we'll have to wait until the marketing page is there?
Don't want any links to a 404 i guess?

PR #445, opened 2019-04-19
APM-307 Explain how to integrate Apple Pay

willemstuursma:


source/wallets/applepay.rst
```diff
@@ -0,0 +1,65 @@
...
+without the ``method`` parameter, the Apple Pay button will be shown in the checkout during method selection:
+
+.. note :: add picture here
+
+Integrating in your own check out
+---------------------------------
+
+At the moment, only limited support for integrating Apple Pay in your own checkout is available.
+
+First, you must indicate to the :doc:`List Method API </reference/v2/methods-api/list-methods>` which wallets you
```
ndijkstra:
```suggestion
First, you must indicate to the :doc:`List Methods API </reference/v2/methods-api/list-methods>` which wallets you
```

source/wallets/applepay.rst
```diff
@@ -0,0 +1,65 @@
...
+<https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_js_api/checking_for_apple_pay_availability>`_.
+
+.. code-block:: javascript
+   :linenos:
+
+   if (window.ApplePaySession && window.ApplePaySession.canMakePayments()) {
+     // Add ApplePay to the method selection in your checkout
+   }
+
+Finally, when the shopper selects Apple Pay from your method selection, you should use the Create Payment API to create
```
ndijkstra:
```suggestion
Finally, when the shopper selects Apple Pay from your method selection, you should use the Payments API to create
```

source/wallets/applepay.rst
```diff
@@ -0,0 +1,65 @@
...
+a payment with the ``method`` parameter set to ``applepay`` and redirect the shopper to the ``_links.checkout`` URL
+returned.
+
+When the shopper pays with Apple pay, a payment of the method ``creditcard`` will be created. We will then call your
```
ndijkstra:
```suggestion
When the shopper pays with Apple Pay, a payment with as method ``creditcard`` will be created. We will then call your
```

source/reference/v2/methods-api/list-methods.rst
```diff
@@ -93,6 +93,18 @@ Parameters
 
        Example: ``https://api.mollie.com/v2/methods?resource=orders&billingCountry=DE``
 
+   * - ``includeWallets``
+
+       .. type:: string
+          :required: false
+
+     - A comma separated list of the wallets you support in your checkout. Wallets often require custom code to check if
```
ndijkstra:
```suggestion
     - A comma-separated list of the wallets you support in your checkout. Wallets often require custom code to check if
```

willemstuursma:
@mattijsbliek can you set me up with a picture of how it looks like?

mattijsbliek:
@willemstuursma What part are you referring to exactly?

willemstuursma:
https://github.com/mollie/api-documentation/pull/445/files#diff-0f5b07bf3433a74c05445e183ad2df66R28 the Apple Pay button in the method selection screen

mattijsbliek:
Will look into it!

In the meantime I’ve added two examples, but I don’t think they’re super clear. Also, I’m wondering whether it wouldn’t be easier to just have a single example. Can you guys take a look and let me know what you think?

/cc @albert-carreras

mattijsbliek:
Just added a picture of a checkout screen with the Apple Pay button.

source/wallets/applepay.rst
```diff
@@ -0,0 +1,213 @@
...
+     <select class="js-select-method">
+       <option value="creditcard">Credit card</option>
+       <option value="ideal">iDEAL</option>
+       <option value="paypal">PayPal</option>
+       <option value="banktransfer">Bank transfer</option>
+     </select>
+   </form>
+
+   <!-- Only load this script if Apple Pay is enabled on your profile. See step 1 above. -->
+   <script src="/path/to/apple-pay.js"></script>
```
willemstuursma: It might help to move the JS into a `<script>` block here instead of two code blocks.

willemstuursma: The check is missing the validation from the Mollie API that Apple Pay is available.

mattijsbliek: Good suggestion, makes the example much clearer 👍 Also addressed the missing check.

source/wallets/applepay.rst
```diff
@@ -0,0 +1,213 @@
...
+   :linenos:
+
+   // apple-pay.js
+   const detectApplePay = () => {
+      if (!window.ApplePaySession || !ApplePaySession.canMakePayments()) {
+        // Apple Pay is not available
+        return;
+      }
+
+      ApplePaySession.canMakePaymentsWithActiveCard('merchant.dev.mollie').then(canMakePayments => {
```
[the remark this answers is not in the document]
mattijsbliek: Discussed this with Albert and rolling with `canMakePayments` only for now. We can always "upgrade" to `canMakePaymentsWithActiveCard` if we receive a lot of pushback from merchants.

The additional benefit of using `canMakePayments` is that it’s faster, so it will create more seamless experience for the consumer.

willemstuursma:
@mattijsbliek can you make the dynamic `<option>` elements be selected by default? Kinda also what Apple recommends.

mattijsbliek:
I actually opted against that, since we’re not checking `canMakePaymentsWithActiveCard`. I wouldn’t set it as the default unless we know for sure that the device can be used to pay with.

source/wallets/applepay.rst
```diff
@@ -0,0 +1,202 @@
...
+       <option value="paypal">PayPal</option>
+       <option value="banktransfer">Bank transfer</option>
+     </select>
+   </form>
+
+   <!-- 
+   Only load this script if Apple Pay is enabled on your Mollie profile. 
+   For example with PHP:
+
+   <?php if ($applePayEnabled) : ?>
```
albert-carreras: Using this check complicates things unnecessarily IMO.
It's also not really necessary, they can add it if they want to but they don't need to. 

We could add it as a recommendation, but I would leave it out and just leave the JS and HTML

source/wallets/applepay.rst
```diff
@@ -0,0 +1,202 @@
...
+     <body>
+       <form>
+         <!-- Other checkout fields like billing address etc. -->
+
+         <ul class="payment-methods">
+           <!-- 
+           Only show this method if Apple Pay is enabled on your Mollie profile. 
+           For example with PHP:
+
+           <?php if ($applePayEnabled) : ?>
```
albert-carreras: Same here... It's not needed.

source/wallets/applepay.rst
```diff
@@ -0,0 +1,202 @@
...
+           </li>
+           ...
+         </ul>
+       </form>
+
+       <!-- 
+       Only load this script if Apple Pay is enabled on your Mollie profile. 
+       For example with PHP:
+
+       <?php if ($applePayEnabled) : ?>
```
albert-carreras: Same here... It's not needed.

source/wallets/applepay.rst
mattijsbliek: But it is necessary right? If Apple Pay is disabled on their Mollie account for any reason, they should not load it.

PR #453, opened 2019-04-30
Add to the webhook guide that we call the webhook for canceling a refund

ndijkstra:


source/changelog/v2/changelog.rst
```diff
@@ -56,6 +56,11 @@ Monday, 25th
 
 - Updated the list of available profile :doc:`merchant category codes </reference/v2/profiles-api/create-profile>`
 
+Wednesday, 6th
```
willemstuursma: This seems to be in the wrong spot.

ndijkstra: This change is already live since February 6th

willemstuursma: Screen Shot 2019-04-30 at 14 53 10

I mean 6th is between 21st and 25th.

ndijkstra: Totally missed that 😅 should be ok now.

PR #454, opened 2019-05-08
APM-346 Add polish złoty to P24

willemstuursma:


source/changelog/v2/changelog.rst
```diff
@@ -6,6 +6,11 @@ changes are documented here.
 May 2019
 ==========
 
+TBD
+---
+
+- Polish złoty is now supported as a currency for the Przelewy24 payment method.
```
ndijkstra: I would add here also the currency code

PR #467, opened 2019-06-11
Get Payment: clarify settlement amount

JorikMollie:
Integrator was confused why the settlementAmount was missing for a payment. Clarified our API documentation.


willemstuursma:
@JorikMollie it is also omitted if the amount is unknown, for example if a payment is created without a method.

PR #471, opened 2019-06-14
Update status flow image

ndijkstra:


lode:
I guess you need this because something changed, but I wanted to give some general feedback that I don't fully understand either of these because of the crossing lines. For example:

- the knot shown between open and paid/canceled/expired suggests that only there you can change direction; such a knot is not shown for going from open to authorized, does that mean it is not possible?
- going from authorized further via the knots suggests you can only go to paid?
- or if I should not interpret it that strict, there are two routes from open to paid

PR #478, opened 2019-06-28
APM-452 allow get settlement via settlement reference

LeonidasMollie:
APM-452 Updated endpoint docs, examples and added entry to the v2 changelog.


willemstuursma:
Don't forget to put the date in, and the new month July probs.

PR #481, opened 2019-07-04
APM-472 Add MyBank to API docs

ndijkstra:


source/changelog/v2/changelog.rst
```diff
@@ -6,6 +6,12 @@ changes are documented here.
 July 2019
 =========
 
+Thursday, 4th
+-------------
+
+- Added ``mybank`` as new payment method. Currently, it's not possible to activate MyBank via the Dashboard.
```
shirshir: Maybe use the capitalised name and put the id in a comment.

### Added MyBank (``mybank``) as new payment method.

source/index.rst
```diff
@@ -49,6 +49,7 @@ Mollie is always adding new payment methods. The Mollie API currently supports t
 * `KBC/CBC Payment Button <https://www.mollie.com/en/payments/kbc-cbc>`_
 * `Klarna Pay later  <https://www.mollie.com/en/payments/klarna-pay-later>`_
 * `Klarna Slice it <https://www.mollie.com/en/payments/klarna-slice-it>`_
+* `MyBank <https://www.mollie.com/en/payments/mybank>`_
```
shirshir: URL does not work.

ndijkstra: We don't have one yet so I'll remove the link

PR #484, opened 2019-07-15
Adding expiryDate parameter to Create Order

MollieRick:


willemstuursma:
What about `expiresAt` instead with a full date and time, to mimic the result from the API?

LeonidasMollie:
> What about `expiresAt` instead with a full date and time, to mimic the result from the API?

I agree that changing the name to `expiresAt` would favour consistency.

MollieRick:
Good point gentlemen! 👍  I will change it.

MollieRick:
Fieldname changed. We decided to leave the time out on submitting. We use the current time like we do when there is no custom expire date provided.

PR #488, opened 2019-08-08
Add "null" as possible method values

martinmollie:
When you create a payment without a method the value is "null". This is also the case if you do a get payment payment call.


ndijkstra:
How about Orders and the V1 API?

martinmollie:
you're right, same with Orders and v1. Can I still make those changes in this PR or do I need to make a new one?

willemstuursma:
@martinmollie keep working in this PR

martinmollie:
@ndijkstra we don't refer to possible values in the Orders API so I left that unchanged.

PR #494, opened 2019-08-19
expiresAt is an optional parameter.

JorikMollie:
Changed the expiresAt parameter to optional


willemstuursma:
Cheers, thanks for adding the explanation as well.

PR #505, opened 2019-09-01
Add time zone explanation to Invoices v1 API

mollierobbert:


dabroek:
Isn't this comment relevant for date output of _all_ API endpoints? Next to that I would use `Europe/Amsterdam` as timezone, since that is the standard.

mollierobbert:
Sure, done. It only applies to `date` fields. For `datetime` fields, we already use UTC notation.

willemstuursma:
I'm not sure the timezone is useful for _dates_ @mollierobbert. Most of the dates are for administration, and without being able to convert a date into another timezone its kinda useless to make it so explicit.

mollierobbert:
Yeah, I was thinking the same thing. Was gonna ask for more opinions tomorrow.

It would be kinda awkward for merchants that are more than 12 hours ahead or behind to see e.g. their subscriptions being fired on a different day than what they anticipated.

We don't currently have merchants that far away though, so I can't imagine it'll be an issue any time soon.

We could also adjust internal processes to interpret these dates as UTC if they ever pose an issue.

willemstuursma:
I would reject this PR then.

mollierobbert:
I agree.

PR #509, opened 2019-09-06
APM-603 - Get Subscription V2 API endpoint should expose _links.payments

kgeorgiev:


source/reference/v2/subscriptions-api/get-subscription.rst
```diff
@@ -226,6 +226,12 @@ Response
               .. type:: URL object
 
             - The URL to the subscription retrieval endpoint documentation.
+          * - ``payments``
```
ndijkstra: Can you place it above the documentation one?

PR #524, opened 2019-10-11
IF-10 Documentation Mollie Components

willemstuursma:


Marinolinderhof:
Could you change the name from elements to components?

willemstuursma:
Thanks @Marinolinderhof

PR #534, opened 2019-10-23
IF-90 renaming to components

Marinolinderhof:


source/guides/mollie-components/reference.rst
```diff
@@ -250,9 +250,9 @@ ES6
     const callback = (event)=> { console.log('We need a real world example here', event.type) }
     cardNumberEl.on('change', callback);
```
willemstuursma: un updated example

willemstuursma: example on line 243 not updated

PR #539, opened 2019-11-05
Update create-payment.rst

TomKriek:
Expanded the example with a value where the thousands separator would be present and show that it is not needed. The value should be a string with 2 decimals use a . as the decimal point and not use a thousands separator.


willemstuursma:
Thanks @TomKriek!

PR #542, opened 2019-11-06
adding the payment param `applePayPaymentToken` to the order endpoint

adirkuhn:
adding the payment param `applePayPaymentToken` to the order endpoint

### DIFF
--- a/source/changelog/v2/changelog.rst
+++ b/source/changelog/v2/changelog.rst
@@ -6,6 +6,11 @@ changes are documented here.
 November 2019
 =============
 
+Thursday, 7th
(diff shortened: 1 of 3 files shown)


willemstuursma:
It would be good to update https://docs.mollie.com/guides/applepay-direct-integration#send-the-apple-pay-payment-token-to-mollie now that the Create Order API can also be used.

willemstuursma:
Can you add it to the changelog:

> The Apple Pay direct integration can now also be used with the Create Order API.

adirkuhn:
I've made the changes

PR #544, opened 2019-11-07
APM-747 intra-eu are cards from the EEA

ndijkstra:


ndijkstra:
Not sure if I should add it to the changelog @willemstuursma

willemstuursma:
No need to update the changelog.

PR #558, opened 2019-11-28
APM-748 allowed shopper country

ndijkstra:


source/reference/v2/payments-api/create-payment.rst
```diff
@@ -181,6 +181,15 @@ Parameters
...
        may be supplied to indicate which of the consumer's accounts should be credited.
 
+   * - ``allowedShopperCountry``
+
+       .. type:: string
+          :required: false
+
+     - For digital goods, you must make sure to apply the VAT rate from your customer’s country in most jurisdictions.
+       Use this parameter to restrict the payment methods available to your customer to methods from a single country only.
+       Set it to the country which’s VAT rates you have used for the order to ensure your customer’s country matches the VAT country.
```
pabloroman: Can we get this double-checked by a native English speaker? I'm not sure whether "Set it to the country _which’s_ VAT rates" is grammatically correct.

sandervanhooft:
This is great!

sandervanhooft:
Would go with `allowedCountry` though.

sandervanhooft:
If I restrict it to `NL`, would a Belgian SEPA card be disallowed?

ndijkstra:
> If I restrict it to `NL`, would a Belgian SEPA card be disallowed?

@sandervanhooft We start supporting it like this:


Country | Methods allowed
-- | --
NL | iDEAL, gift cards
Belgium | Bancontact, ING Home’pay, KBC and Belfius
Germany | Giropay
Austria | eps
Poland | Przelewy24

We will improve this so we can add more payment methods like Credit Card.

sandervanhooft:
@nick thanks. Can this way of restricting be used as proof for MOSS regulations?

I think it could be, but I'm not sure tbh.

source/reference/v2/orders-api/get-order.rst
```diff
@@ -131,6 +131,13 @@ Response
 
      - The person and the address the order is billed to. See below.
 
+   * - ``shopperCountryMustMatchBillingCountry``
+
+       .. type:: boolean
+
+     - If set to true during order creation, we restrict the payment methods available to your customer
+       to methods from a single country only.
```
pabloroman: Nitpick: can you make this consistent with the description of `create-order.rst`? "to methods from a single country only." --> "to methods from the billing country only."

source/reference/v2/payments-api/create-payment.rst
sandervanhooft: @pabloroman How about something like this...

Also, do we need a note explaining the difference with `locale` somewhere?

The EU has special tax regulations (MOSS) for digital goods. These regulations require you to apply the customer country's VAT rate to these products and to keep two records of proof that the right VAT rate was applied.

The customer's billing address can be used as one piece of proof. By restricting the payment methods to the billing address country the used payment method can be used as another piece of proof. This can be achieved using the `restrictShopperCountry` parameter.

PR #560, opened 2019-12-02
PL-1098 Add changelog entry about new IP addresses.

hvalle:
Added an entry to the changelog to mention the use of new IP's to send webhooks. It is noted that users should not verify these IP's anyway.


Daanvm:
There's also a v2 changelog file that should include the same changes.

PR #561, opened 2019-12-03
APM-808 - Adding info for the 'extraMerchantData' for Klarna.

MollieRick:


source/reference/v2/payments-api/create-payment.rst
```diff
@@ -438,6 +438,24 @@ KBC/CBC Payment Button
...
+   :widths: auto
+
+   * - ``extraMerchantData``
+
+       .. type:: string
+          :required: false
+
+     - There is additional purchase information required for some industries. You can submit your
+       extra data in this field if you have this agreed with Klarna. This field should be a string
+       containing any of the allowed keys and sub objects described at the
```
ndijkstra: Shouldn't we mention JSON somewhere? Because it's a JSON string, right?

source/reference/v2/payments-api/create-payment.rst
```diff
@@ -447,6 +447,25 @@ KBC/CBC Payment Button
...
+
+.. list-table::
+   :widths: auto
+
+   * - ``extraMerchantData``
+
+       .. type:: string
+          :required: false
+
+     - There is additional purchase information required for some industries. You can submit your
```
willemstuursma: I thought merchants would send it as objects and we would serialize it for them @MollieRick ?

MollieRick: Please see #562

PR #565, opened 2019-12-04
Add documentation on testing

willemstuursma:


source/guides/mollie-components/testing.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Then, you can implement the remainder of Mollie Components as specified in :doc:`our guide <overview>`. You can use the
+**Test API key** instead of the Live API key when :doc:`creating the Payment </reference/v2/payments-api/create-payment>`.
+
+Card numbers for testing
+------------------------
+
+You can use these cards for testing, with any CVV or expiry date.
+
++------------------+---------------------+
```
mvdpanne: In order to allow for testing error conditions for both 3-D Secure and non 3-D Secure payments, we need two test credit card numbers per brand.

The last four digits will then be used to determine which path to follow in the code. So the last four digits of the cards in the 3-D Secure group cannot overlap with the last four digits of the cards in the non 3-D Secure group.

source/guides/mollie-components/testing.rst
```diff
@@ -0,0 +1,64 @@
...
+
+   var mollie = Mollie('pfl_3RkSN1zuPE', { locale: 'nl_NL', testmode: true });
+
+Then, you can implement the remainder of Mollie Components as specified in :doc:`our guide <overview>`. You can use the
+**Test API key** instead of the Live API key when :doc:`creating the Payment </reference/v2/payments-api/create-payment>`.
+
+Card numbers for testing
+------------------------
+
+You can use these cards for testing, with any CVV or expiry date.
```
mvdpanne: NBC: I would think 'and' fits better than 'or' here.

source/theme/styles/components/_parameters.scss
```diff
@@ -61,6 +61,14 @@
   text-align: right;
 }
 
+.rst-content .section table > tbody > tr > td.right-align {
```
willemstuursma: I tried adding this @mattijsbliek but I couldn't get it to work locally if I made changes and rebuilt the CSS files. Can you help me out?

mattijsbliek: I’m completely packed today. Maybe @Marinolinderhof can help out? Otherwise you’re first on my list tomorrow.

PR #566, opened 2019-12-05
Mollie Components: Clarify error handling examples on the client-side

pabloroman:
Ideally we should add more information about the different error messages we return @mattijsbliek . I can add those if you give me a list.

### DIFF
--- a/source/guides/mollie-components/overview.rst
+++ b/source/guides/mollie-components/overview.rst
@@ -159,6 +159,11 @@ You can then place the ``cardToken`` in a hidden input to submit it to your back
 
      const { token, error } = await mollie.createToken();
 
+     if (error) {
+       // Something wrong happened while creating the token. Handle this situation gracefully.
+       return;
+     }
+
      // Add token to the form
      const tokenInput = document.createElement('input');
      tokenInput.setAttribute('type', 'hidden');
@@ -167,7 +172,7 @@ You can then place the ``cardToken`` in a hidden input to submit it to your back
 
      form.appendChild(tokenInput);
 
-     // Re-submit form to the server
+     // Submit form to the server
      form.submit();
    });
 


source/guides/mollie-components/overview.rst
```diff
@@ -159,16 +159,20 @@ You can then place the ``cardToken`` in a hidden input to submit it to your back
 
      const { token, error } = await mollie.createToken();
 
-     // Add token to the form
-     const tokenInput = document.createElement('input');
-     tokenInput.setAttribute('type', 'hidden');
-     tokenInput.setAttribute('name', 'cardToken');
-     tokenInput.setAttribute('value', token);
+     if (error) {
+       // Something wrong happened while creating the token. Handle this situation gracefully.
```
mattijsbliek: Can you early return so you don’t need the `else`? All good otherwise 🙌

willemstuursma:
An example thing that can happen is: the card is not supported (e.g. Discover card). This should be reflected back to the shopper, maybe even automatically.

source/guides/mollie-components/overview.rst
```diff
@@ -159,16 +159,21 @@ You can then place the ``cardToken`` in a hidden input to submit it to your back
...
+       return;
+     } else {
```
mattijsbliek: You can now get rid of this `else` statement :)

mattijsbliek:
> An example thing that can happen is: the card is not supported (e.g. Discover card). This should be reflected back to the shopper, maybe even automatically.

This is already handled automatically by Mollie Components. The only thing we don’t handle is Mollie supporting the card, but it not being enabled on the merchant’s account (e.g. Amex being turned off).

source/guides/mollie-components/overview.rst
pabloroman: Shame on me 😱 Sorted!

mattijsbliek: For sure!

PR #568, opened 2019-12-09
Update styling.rst

Marinolinderhof:
Make it a bit more clear how to style Mollie components.


source/guides/mollie-components/styling.rst
```diff
@@ -4,16 +4,17 @@ Styling Mollie Components
 :doc:`Mollie Components <overview>` can be fully styled to ensure a smooth blend-in with your checkout.
 
 Most of the styling should be applied to the ``.mollie-component`` container instead of passing styling to the
```
mattijsbliek: Should we maybe lead with explaining that will add this div to the mounting target of the element? Maybe even some HTML? What do you think?

MollieRick:
Can we merge this?
It's open for a while now.

PR #582, opened 2020-01-14
APM-944 Add interval to update subscription API

ndijkstra:


sandervanhooft:
https://github.com/mollie/mollie-api-php/pull/435

sandervanhooft:
Just to be sure: will sending an unchanged interval NOT trigger an exception?

ndijkstra:
> Just to be sure: will sending an unchanged interval NOT trigger an exception?

No, this should not be a problem

sandervanhooft:
Great @ndijkstra !

PR #589, opened 2020-01-30
PZ-2674 - Adding shipments as embed possibility to the orders API

MollieRick:


willemstuursma:
No changelog?

MollieRick:
Will add it when I have a date :)

PR #597, opened 2020-03-03
WIP: PZ-2834 Google Pay documentation

willemstuursma:


mollierobbert:
Gonna reject this PR for now since it is collecting dust and conflicts as long as we don't actually support Google Pay.

PR #598, opened 2020-03-05
PZ-84 Duplicate refunds are now blocked by the API

willemstuursma:


source/reference/v2/refunds-api/create-refund.rst
```diff
@@ -198,3 +198,24 @@ Response
...
        }
    }
+
+Response (duplicate refund detected)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: http
+   :linenos:
+
+   HTTP/1.1 429 Too Many Requests
```
lode: Why was this status code chosen? From a client perspective I interpreted this as 'you made too many calls in a close time, try it again after a while. Preferable combined with a header indicating when I can try again.

However, as I understand the feature, you shouldn't try again at all because you already refunded. I know you can refund multiple times, but that is not a very common scenario right?

Maybe `409 Conflict` or `425 Too Early` is a better one?

willemstuursma: Thanks, that is an excellent suggestion. We'll change it to `409 Conflict`.

PR #607, opened 2020-03-23
FFS-2976 Reflect invoicing changes in settlements API docs"

Thijs-Riezebeek:


source/reference/v2/settlements-api/get-settlement.rst
```diff
@@ -222,12 +223,25 @@ Response
 
                    - The gross total costs for this payment method (includes VAT).
 
+          * - ``invoiceId``
+
+              .. type:: object
```
mollierobbert:
```suggestion
              .. type:: string
```

source/changelog/v2/changelog.rst
```diff
@@ -6,6 +6,17 @@ changes are documented here.
 March 2020
 ==========
 
+Monday, 23rd
+------------
+
+- It is now possible to link all costs in a settlement to all invoices associated with it.
```
shuisman: Maybe extend this a bit more with a reason

> Mollie is moving to a monthly invoicing system. As some settlements span multiple months, we have added an invoiceId to each period in a settlement.

source/reference/v2/settlements-api/get-settlement.rst
```diff
@@ -222,12 +223,25 @@ Response
...
+
+            - The ID of the invoice that was created to invoice specifically the costs in this
+              month/period.
+
+              If an individual month/period has not been invoiced yet, then this field will not
+              be present until that invoice is created.
+
    * - ``invoiceId``
```
shuisman: Should we make this deprecated to stimulate merchants / partners using the new invoiceId in the period object?

source/changelog/v2/changelog.rst
mollierobbert: Great suggestion!

Thijs-Riezebeek: Incorporated the suggestion and reworded it a little bit to go with the rest.

source/reference/v2/settlements-api/get-settlement.rst
Thijs-Riezebeek: I think we should, but we'll have to wait until we actually can!

Thijs-Riezebeek: So now that we can ,this is being picked up. Added a warning and mentioned the deprecation.

PR #615, opened 2020-04-16
Update testing.rst

iXeQuta:
clarified when and/or how the webhook is called when in test-mode

### DIFF
--- a/source/guides/testing.rst
+++ b/source/guides/testing.rst
@@ -27,6 +27,10 @@ For paid test mode payments the resource will also include the ``changePaymentSt
 :doc:`create a refund </payments/refunds>` or chargeback for that payment directly from our hosted payment page.
 This can be used to test refund and chargeback functionality.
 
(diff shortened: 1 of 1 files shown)


source/guides/testing.rst
```diff
@@ -27,6 +27,10 @@ For paid test mode payments the resource will also include the ``changePaymentSt
 :doc:`create a refund </payments/refunds>` or chargeback for that payment directly from our hosted payment page.
 This can be used to test refund and chargeback functionality.
 
+If you are using the Mollie checkout the status failed or cancelled will not trigger the webhook call and redirect URL. Instead the shopper is returned to the payment method selection page. The shopper can select an other payment method to try again.
```
MollieRick, line 30: This not true. If you specify a method (instead of leaving it blank) it will trigger a webhook and redirect.

MollieRick (changes requested):

PR #617, opened 2020-04-22
PZ-2667 Document credit card mandate export procedure for competing PSPs

willemstuursma:


source/payments/importing-mandates.rst
```diff
@@ -0,0 +1,89 @@
+Importing credit card mandates from Mollie
+==========================================
+
+.. note:: This document is aimed at competing payment service providers who wish to import existing
+          credit card mandates from Mollie for a merchant.
+
+In the event that a merchant wishes to switch to a different payment service provider, the merchant
+can export a list of his / her credit card mandates from the Mollie Dashboard.
+
+You must provide the merchant with the export format you would like to receive and a public `OpenPGP <https://www.openpgp.org/>`_
```
pabloroman, line 10: What are the different export formats we provide?

source/payments/importing-mandates.rst
```diff
@@ -0,0 +1,89 @@
...
+key the export should be encrypted to.
+
+The merchant must contact Mollie to enable the functionality to export the mandates.
+
+Then, the merchant can upload your public key in the Mollie Dashboard and export the mandates from
+there. Additionally, a public key of Mollie must be downloaded.
```
pabloroman, line 16: Where can this public key be found? also in the Dashboard?

willemstuursma: Yes.

source/payments/importing-mandates.rst
willemstuursma: I'll clarify. I don't want to name any competitors.

source/payments/importing-mandates.rst
```diff
@@ -0,0 +1,89 @@
+Importing credit card mandates from Mollie
```
sandervanhooft, line 1: Shouldn't the title be "Exporting" instead of "Importing"?

sandervanhooft: If so, the filename should be `exporting-mandates.rst`

willemstuursma: Hi Sander. This is for employees of competing PSPs that wish to import mandates from Mollie. So from their pov it is importing as it mostly deals with the decryption and signature validation.

We will provide public documentation on exporting mandates from competing PSPs later.

sandervanhooft:
Can this also be used for moving mandates between Mollie accounts? Could be valuable for profile carve-outs (I.e. a profile that becomes a separate business entity)

willemstuursma:
Hi @sandervanhooft, that is an interesting idea. I'll come back to you on that.

PR #619, opened 2020-04-23
Add some Python code blocks

J0XX0:
Hi Mollie,

I've added a couple of Python code blocks I thought would make the docs a little bit more clear. 

Hope it's sufficient, I might create another pull request with new functions when we need them.


willemstuursma:
Cool, thanks!

PR #620, opened 2020-04-28
Adds status to methods and adds enablement of all methods

albert-carreras:


source/reference/v2/methods-api/get-method.rst
```diff
@@ -156,6 +156,12 @@ Response
             - The URL for a payment method icon in vector format. Usage of this format is preferred since it can scale
               to any desired size.
 
+   * - ``status``
+
+       .. type:: string
+
+     - The status that the method is in. Possible values: ``activated`` ``disabled`` ``pending-boarding`` ``pending-review`` ``pending-external`` ``rejected``
```
mattijsbliek, line 163: I think it would be good to explain what every status means.

source/reference/v2/profiles-api/enable-method.rst
```diff
@@ -23,8 +23,8 @@ Enable payment method
 
 Enable a payment method on a specific or authenticated profile to use it with payments.
 
-.. note:: Not all payment methods can be enabled via this API call. The API will return an error when this is the case
-          with a link to the Mollie Dashboard where the method can be enabled manually.
+.. note:: Some payment methods might need extra steps to be enabled (like PayPal and its authentication).
+          In those cases, the status will be set to pending and the response will contain a link to continue the enablement.
```
mattijsbliek, line 27: `pending-external`? Are you just describing PayPal here, or every scenario that can yield a `pending-*` status?

source/reference/v2/profiles-api/enable-method.rst
```diff
@@ -120,7 +121,7 @@ Request (method that can't be enabled)
       $profile = $mollie->profiles->get('pfl_v9hTwCvYqw');
 
       try {
```
mattijsbliek, line 123: Do we still throw? I don't think so right?

source/reference/v2/methods-api/get-method.rst
```diff
@@ -156,6 +156,51 @@ Response
...
+
+       .. list-table::
+          :widths: auto
+
+          * - ``activated``
+
+              .. type:: string
+
+            - The payment method is enabled.
```
mattijsbliek, line 172:
```suggestion
            - The payment method is activated and ready for use.
```

source/reference/v2/methods-api/get-method.rst
```diff
@@ -156,6 +156,51 @@ Response
...
+
+          * - ``disabled``
+
+              .. type:: string
+
+            - The payment method is disabled.
```
mattijsbliek, line 178: I think this might be confusing. If I read the code:
```
        // If the method is disabled and not rejected we don't return a status
        if (!$profile->isMethodEnabled($spec->getPaymentType()->getPaymentTypeId())) {
            return self::STATUS_DISABLED;
        }
```

it just means that the payment method is not requested. I'm wondering if we shouldn't return `null` here to be honest. But that would also require a change to the code. Maybe good for discussion with the team. We can only merge this once and after that we're stuck with it forever.

source/reference/v2/methods-api/get-method.rst
```diff
@@ -156,6 +156,51 @@ Response
...
+
+          * - ``pending-boarding``
+
+              .. type:: string
+
+            - The payment method has been enabled but needs the merchant to finish the boarding process to be active.
```
mattijsbliek, line 184:
```suggestion
            - Mollie is waiting for you to finish onboarding in the Merchant Dashboard before the payment method can be activated.
```

source/reference/v2/methods-api/get-method.rst
```diff
@@ -156,6 +156,51 @@ Response
...
+
+          * - ``pending-review``
+
+              .. type:: string
+
+            - The payment method has been enabled but needs a review from Mollie to be active.
```
mattijsbliek, line 190:
```suggestion
            - Mollie needs to review your request for this payment method before it can be activated.
```

source/reference/v2/methods-api/get-method.rst
```diff
@@ -156,6 +156,51 @@ Response
...
+
+          * - ``pending-external``
+
+              .. type:: string
+
+            - The payment method has been enabled but needs the an external process to be active.
```
mattijsbliek, line 196:
```suggestion
            - Activation of this payment method relies on you taking action with an external party, for example signing up with PayPal or a giftcard issuer.
```

source/reference/v2/methods-api/get-method.rst
```diff
@@ -156,6 +156,51 @@ Response
...
+
+          * - ``rejected``
+
+              .. type:: string
+
+            - The payment method has been rejected by Mollie.
```
mattijsbliek, line 202:
```suggestion
            - Your request for this payment method was rejected. Whenever Mollie rejects such a request, you will always be informed via email.
```

source/reference/v2/profiles-api/enable-method.rst
```diff
@@ -23,8 +23,8 @@ Enable payment method
 
 Enable a payment method on a specific or authenticated profile to use it with payments.
 
-.. note:: Not all payment methods can be enabled via this API call. The API will return an error when this is the case
-          with a link to the Mollie Dashboard where the method can be enabled manually.
+.. note:: Some payment methods might need extra steps to be enabled (like PayPal and its authentication).
```
mattijsbliek, line 26:
```suggestion
.. note:: Some payment methods might need extra steps to be activated, for example an OAuth connection with the supplier.
```

source/reference/v2/profiles-api/enable-method.rst
```diff
@@ -23,8 +23,8 @@ Enable payment method
...
+          In those cases, the status will be set to ``pending-external`` and the response will contain a link to continue the enablement.
```
mattijsbliek, line 27:
```suggestion
          In those cases, the status will be set to ``pending-external`` and the response will contain a link to continue the activation.
```

source/reference/v2/profiles-api/enable-method.rst
```diff
@@ -39,8 +39,8 @@ An objects of ``method`` will be returned as described in :doc:`Get method </ref
 Example
 -------
 
-Request (method that can be enabled)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Request (method that can be immediately enabled)
```
mattijsbliek, line 42: Should we have two examples? Since this endpoint no longer throws, the response is the same. Also the response on line 95 is incorrect since it should contain all the details as well. My suggestion would be to just stick with one Request/Response example.

source/reference/v2/methods-api/get-method.rst
albert-carreras: I just checked and we don't really use that status for anything yet, so a change would be easy.

source/reference/v2/profiles-api/enable-method.rst
albert-carreras: Fair point, will merge them

source/reference/v2/profiles-api/enable-method.rst
```diff
@@ -78,79 +68,31 @@ Response
...
+            "size2x": "https://www.mollie.com/external/icons/payment-methods/ideal%402x.png",
+            "svg": "https://www.mollie.com/external/icons/payment-methods/ideal.svg"
+        },
+        "status": "activated",
+        "_links": {
+            "self": {
+                "href": "https://api.mollie.com/v2/methods/ideal",
+                "type": "application/hal+json"
             },
             "documentation": {
```
lvgunst, line 93: This should not be under the `documentation` key?

albert-carreras: What do you mean? The response comes from postman.

PR #623, opened 2020-05-06
Change warning on top of page

JustinSangers:
Instead of saying it's still possible to use the Reseller API, we now want to communicate that it is only available for selected partners.

### DIFF
--- a/source/reference/reseller-api/guides/overview.rst
+++ b/source/reference/reseller-api/guides/overview.rst
@@ -1,8 +1,7 @@
 Reseller API
 ============
 
-.. warning:: The Reseller API has been deprecated and can still be used for the foreseeable future. We recommend to
-             change to the Mollie v2 API.
+.. warning:: The Reseller API has been deprecated. Only selected partners still have access to this legacy functionality. As an     alternative, we recommend using :doc:`Mollie Connect </oauth/overview>` and the :doc:`Mollie v2 API </reference/v2>`.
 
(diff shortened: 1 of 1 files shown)


source/reference/reseller-api/guides/overview.rst
```diff
@@ -1,8 +1,7 @@
 Reseller API
 ============
 
-.. warning:: The Reseller API has been deprecated and can still be used for the foreseeable future. We recommend to
-             change to the Mollie v2 API.
+.. warning:: The Reseller API has been deprecated. Only selected partners still have access to this legacy functionality. As an     alternative, we recommend using :doc:`Mollie Connect </oauth/overview>` and the :doc:`Mollie v2 API </reference/v2>`.
```
vernondegoede, line 4: A top-level page `reference/v2` does not exist.

Please link to the first page of the reference instead: `reference/v2/payments-api/create-payment`

vernondegoede: Also, there are too many spaces between `an` and `alternative`

vernondegoede, line 4:
```suggestion
.. warning:: The Reseller API has been deprecated. Only selected partners still have access to this legacy functionality. As an alternative, we recommend using :doc:`Mollie Connect </oauth/overview>` and the :doc:`Mollie v2 API </reference/v2/payments-api/create-payment>`.
```

PR #626, opened 2020-05-13
Documentation for Gift Card Issuer endpoints. 

actuallymab:


mattijsbliek:
@thiagocordeiro Would be great if you could take the lead on this. I can do a double check if needed :)

source/reference/v2/profiles-api/disable-gift-card-issuer.rst
```diff
@@ -0,0 +1,52 @@
...
+   :url: https://api.mollie.com/v2/profiles/*id*/methods/giftcard/issuers/*issuer*
+
+.. authentication::
+   :api_keys: false
+   :oauth: true
+   :organization_access_tokens: true
+
+.. endpoint::
+   :method: DELETE
+   :url: https://api.mollie.com/v2/profiles/me/methods/giftcard/issuers/*issuer*
```
willemstuursma, line 17: profiles/me should work with an API key right?

actuallymab: it's already there 🙌 https://github.com/mollie/api-documentation/pull/626/files#diff-4549636fa01f88e9699212ddfec29b27R20

source/reference/v2/profiles-api/enable-gift-card-issuer.rst
```diff
@@ -0,0 +1,132 @@
...
+``201`` ``application/hal+json``
+
+..  list-table::
+    :widths: auto
+
+    * - ``resource``
+
+        .. type:: string
+
+      - Indicates the response contains an issuer object. Will always contain ``issuers`` for this endpoint.
```
mattijsbliek, line 43: We talked about this one before, but since all Mollie's resources are singular, shouldn't this be `issuer`?

actuallymab: Well, to be sure I double-checked my feature PR, it was "issuers" before, and I didn't change that part. But we can update it if you want?

mattijsbliek: We probably should right? Would be a bit strange to have singular everywhere but here 🤔

PR #627, opened 2020-05-19
PAA-2430 Add the `lang` param to the Authorize endpoint

dmvdbrugge:
Text itself is proofread and approved by Justin

### DIFF
--- a/source/reference/oauth2/authorize.rst
+++ b/source/reference/oauth2/authorize.rst
@@ -74,6 +74,19 @@ Parameters
 
        Possible values: ``auto`` ``force``
 
+   * - ``lang``
+
+       .. type:: string
+          :required: false
+
+     - Allows you to preset the language to be used in the login / sign up / authorize flow if the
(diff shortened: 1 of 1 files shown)


source/reference/oauth2/authorize.rst
```diff
@@ -74,6 +74,19 @@ Parameters
 
        Possible values: ``auto`` ``force``
 
+   * - ``lang``
```
ndijkstra, line 77: Isn't it better to name this parameter `locale` which we have in the other endpoints as well?

willemstuursma: Yes that's much better @vernondegoede .

willemstuursma:
Please rename to `locale` to align with all other APIs.

source/reference/oauth2/authorize.rst
vernondegoede: That was one of the requirements initially, but since this is not an API call (and Mollie Platform works with `lang` we decided to use `lang` here too).

PR #631, opened 2020-06-08
CC-551 card_declined failure reason

kgeorgiev:


source/reference/v2/payments-api/get-payment.rst
```diff
@@ -788,7 +788,7 @@ Credit card
 
               Possible values: ``authentication_failed``  ``card_expired`` ``inactive_card`` ``insufficient_funds``
               ``invalid_card_holder_name`` ``invalid_card_number`` ``invalid_card_type`` ``invalid_cvv``
-              ``possible_fraud`` ``refused_by_issuer`` ``unknown_reason``
+              ``possible_fraud`` ``refused_by_issuer`` ``card_declined`` ``unknown_reason``
```
mvdpanne, line 791: I think the values should be alphabetically sorted here.

willemstuursma:
What's the difference between `card_declined` and `refused_by_issuer`?

mvdpanne:
@willemstuursma Bart wanted another generic reason that does not mention the issuer, as not all cases relate to the issuer.

He preferred this over using the `unknown_reason`, as the message that is displayed for that failure reason indicates that we don't know why the card was declined and that the shopper should contact their bank for more information. That's not what Bart wanted us to convey.

source/reference/v2/payments-api/get-payment.rst
mollierobbert: Done.

PR #633, opened 2020-06-11
Do not use fonts from different services, instead use local fonts

lvgunst:
We made some changes to the marketing website. During this change we moved the fonts to a different folder. After deploying we got questions why the fonts were broken in different parts of Mollie. This shows how fragile this is and that we shouldn't use fonts from different Mollie service.

This PR copies the font files from the marketing website to a local folder and use those.


source/_static/404.css
```diff
@@ -1,22 +1,22 @@
 @font-face {
   font-family: "Calibre";
   font-weight: 200;
-  src: url("https://cdn.mollie.com/assets/CalibreWeb-Light.woff2") format("woff2"),
-    url("https://cdn.mollie.com/assets/CalibreWeb-Light.woff") format("woff");
+  src: url("./fonts/CalibreWeb-Light.woff2") format("woff2"),
```
vernondegoede, line 4: I think this font should be replaced by Circular (and its source files can then be removed).

PR #635, opened 2020-06-15
Clarify QR Code generation

sandervanhooft:
I've tested this a bit, generating a QR code seems to require explicitly setting the payment method in the payload (i.e. `"method" : "ideal"`).

The docs are not clear on this.


willemstuursma:
Cheers @sandervanhooft.

PR #648, opened 2020-08-11
PZ-3602 Allow issuer to be updated for Payment

feyyazesat:


willemstuursma:
I think it is fine to only mention that the `issuer` field can be updated and refer to the other documentation pages with a deep link.

willemstuursma:
Please add it to the public change log.

package.json
```diff
@@ -37,7 +37,8 @@
     "lint-staged": "^10.0.3",
     "node-sass": "^4.14.1",
     "parcel-bundler": "^1.12.3",
-    "prettier": "^2.0.0"
+    "prettier": "^2.0.0",
+    "sass": "^1.26.10"
```
vernondegoede, line 41: @feyyazesat You don't need this dependency. We already have `node-sass`. Can you remove it again?

PR #662, opened 2020-09-14
PZ-3375 - Adding failureReason for Bancontact

MollieRick:


source/reference/v2/payments-api/get-payment.rst
```diff
@@ -598,6 +598,12 @@ Bancontact
 
             - Only available if the payment is completed – The consumer's bank's BIC / SWIFT code.
 
+          * - ``failureReason``
+
+              .. type:: string
+
+            - Only available if the payment is failed – The (translated) reason why the payment did not succeed.
```
ndijkstra, line 605: If the payment has failed, it can still occur that there is no failureReason.

ndijkstra (approved):

PR #676, opened 2020-10-07
PZ-4014 - Adding more details for vouchers

MollieRick:


source/reference/v1/payments-api/get-payment.rst
```diff
@@ -994,6 +994,63 @@ SOFORT Banking
...
+
+                 * - ``amount``
+
+                     .. type:: decimal
+
+                   - The amount in EUR that was paid with this voucher.
+
+          * - ``remainderAmount``
+
+              .. type:: decimal
```
willemstuursma, line 1042: this should be an amount object.

source/reference/v1/payments-api/get-payment.rst
```diff
@@ -994,6 +994,63 @@ SOFORT Banking
...
+
+                 * - ``issuer``
+
+                     .. type:: string
+
+                   - The ID of the voucher brand that was used during the payment.
+
+                 * - ``amount``
+
+                     .. type:: decimal
```
willemstuursma, line 1036: this should also be an amount object

PR #678, opened 2020-10-09
url to href

jvink:


willemstuursma:
Hi @jvink, thanks for taking the time to fix this!

PR #682, opened 2020-10-16
Update recurring.rst for subscription case

Shutsen:
At the moment it's a bit unclear how to handle a subscription model while having to make a first-time payment. The best solution is to charge the customer the amount that will be periodically charged right away. Next, you can set up a subscription which starts after the first period (see subscription API, startDate)


mollierobbert:
Thanks for these suggestions, Geert. Looks good. 👍 

I iterated on your copy a bit, since the first instalment is often a slightly different amount than the follow-up instalments.

PR #696, opened 2020-11-09
Add details on testing app fees

arnetm:
Testing is not possible with same account as application owner account


arnetm:
@vernondegoede Can you review?

willemstuursma:
It is correct but I would like to rephrase it a bit, will be back later today.

PR #711, opened 2020-12-13
Updated the use of the javascript example.

galstyangroup:
This ID should not be passed in the body but as an additional parameter. Also added metadata example.

### DIFF
--- a/source/reference/v2/customers-api/update-customer.rst
+++ b/source/reference/v2/customers-api/update-customer.rst
@@ -123,7 +123,7 @@ Example
       const mollieClient = createMollieClient({ apiKey: 'test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM' });
 
       (async () => {
-        const customer = await mollieClient.customers.update('cst_8wmqcHMN4U(customerID)' , {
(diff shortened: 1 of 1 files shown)


source/reference/v2/customers-api/update-customer.rst
```diff
@@ -123,10 +123,12 @@ Example
       const mollieClient = createMollieClient({ apiKey: 'test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM' });
 
       (async () => {
-        const customer = await mollieClient.customers.update({
-          customerId: 'cst_8wmqcHMN4U',
+        const customer = await mollieClient.customers.update('cst_8wmqcHMN4U(customerID)' , {
```
vernondegoede, line 126:
```suggestion
        const customer = await mollieClient.customers.update('cst_8wmqcHMN4U' , {
```

vernondegoede (changes requested):

mollierobbert:
Nice catch Harut.

Showing how to update the `metadata` is a nice suggestion as well, but I left it out for now since it would make the JS snippet inconsistent with the other examples.

I'll merge the PR. Thanks!

vernondegoede (approved):

PR #717, opened 2021-01-13
Clarify use of Refresh Token

Timollie:
Based on feedback from partners clarified use of refresh and access tokens.


mollierobbert:
Looks good! Thanks for these changes. I merged them.

PR #727, opened 2021-01-26
Add `year` & `month` parameter to clients-list endpoint

robindirksen1:
For the endpoint `https://api.mollie.com/v2/clients`, it is possible to add extra parameters to the request. 

For example:
- year: 2020 (`?year=2020`)
- month: 12 (`?month=12`)


mollierobbert:
Good catch! Were these parameters hidden for a reason, @vernondegoede? Either way, they are public now :-).

mollierobbert:
Thanks Robin!

PR #730, opened 2021-02-03
add python example for delete mandate

matteobaldelli:


JorikMollie:
I'm checking with the Python developer if this is the canonical way. Will update when I know more.

JorikMollie:
This is the canonical way. @ndijkstra can you merge this?

PR #731, opened 2021-02-03
Label `locale` as optional in `/payments/*` response

rvdsteege:
Responses from the `/v2/payments/*` endpoint do not always include a `locale` (for example when using Apple Pay in Mollie Checkout and no locale was specified on payment creation), therefore `locale` should have the "Optional" label in the documentation.


mollierobbert:
Good catch Reüel, thanks for the edit!

PR #735, opened 2021-02-25
Fix indentation in recurring payments doc

rolfvandekrol:
Because the second line of a ordered list item is missing indentation, it is not recognized as an ordered list item, which result in the `#.` to be included in the output on https://docs.mollie.com/payments/recurring. This pull request should fix that.


willemstuursma:
Cheers.

PR #736, opened 2021-03-01
Update create-payment.rst

arnetm:
Add instructions on Paypal description to avoid "PayPal error code 10412: Payment has already been made for this InvoiceID."


rosstuck (approved):

sandervanhooft:
Perhaps put it in a _note box_? This is a critical piece of information, but still easily overlooked this way.

PR #779, opened 2021-07-12
Update create-order.rst

angelique-mollie:
Remove "Direct debit" at "method" of the orders API, because it cannot be shown in the payment screen.

### DIFF
--- a/source/reference/v2/orders-api/create-order.rst
+++ b/source/reference/v2/orders-api/create-order.rst
@@ -146,7 +146,7 @@ Parameters
        screen but will only show the methods specified in the array. For example, you can use this functionality to only
        show payment methods from a specific country to your customer ``['bancontact', 'belfius']``.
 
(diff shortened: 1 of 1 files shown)


sandervanhooft:
@angelique-mollie I was under the impression that Orders can be used in a recurring setup, meaning that there's still a use case for `directdebit` here. In that case, maybe only the description could use an update, not the list of potential values?

source/reference/v2/orders-api/create-order.rst
```diff
@@ -146,7 +146,7 @@ Parameters
        screen but will only show the methods specified in the array. For example, you can use this functionality to only
        show payment methods from a specific country to your customer ``['bancontact', 'belfius']``.
 
-       Possible values: ``applepay`` ``bancontact`` ``banktransfer`` ``belfius`` ``creditcard`` ``directdebit`` ``eps``
```
MollieRick, line 149: Its correct that it can not be shown on the method screen, but it is possible to use `directdebit` as method for an order.

MollieRick (changes requested):

PR #805, opened 2021-09-14
PZ-4639 Improve PayPal info for description param

ndijkstra:
Schermafbeelding 2021-09-14 om 08 52 58

### DIFF
--- a/source/reference/v2/payments-api/create-payment.rst
+++ b/source/reference/v2/payments-api/create-payment.rst
@@ -419,9 +419,19 @@ PayPal
    :type: string
    :condition: optional
 
-   If a description in the form ``Order <order number>`` is used, the order number is passed to PayPal as the
-   *invoice reference*. This field is searchable in the PayPal merchant dashboard. This field must be unique and should
-   not contain symbols. Alternatively, we will recognize the following keywords:
+   If a description in the form ``Order <order number>`` is used and the <order number> is right after the ``Order``
+   keyword then it is passed to PayPal as the *invoice reference*. This field is searchable in the PayPal merchant
+   dashboard. Also note that the <order number> should be unique across all transactions in PayPal and should not
+   contain symbols.
+
+   For example:
+
+   * ``Order Best Service ABS123`` does not match, no invoice reference is sent to PayPal.
+
+   * ``Best Service Order ABS123`` will match, so "Order ABS123" is sent to PayPal as invoice reference number.
+
+
+   Alternatively, we will recognize the following keywords:
 
    * Cart
    * Order
--- a/source/reference/v2/payments-api/create-payment.rst
+++ b/source/reference/v2/payments-api/create-payment.rst
@@ -419,16 +419,19 @@ PayPal
    :type: string
    :condition: optional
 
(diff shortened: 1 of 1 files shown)


source/reference/v2/payments-api/create-payment.rst
```diff
@@ -419,9 +419,19 @@ PayPal
    :type: string
    :condition: optional
 
-   If a description in the form ``Order <order number>`` is used, the order number is passed to PayPal as the
-   *invoice reference*. This field is searchable in the PayPal merchant dashboard. This field must be unique and should
-   not contain symbols. Alternatively, we will recognize the following keywords:
+   If a description in the form ``Order <order number>`` is used and the <order number> is right after the ``Order``
+   keyword then it is passed to PayPal as the *invoice reference*. This field is searchable in the PayPal merchant
+   dashboard. Also note that the <order number> should be unique across all transactions in PayPal and should not
+   contain symbols.
```
Timollie, line 425: Can you define symbols here? I assume you meant special characters like "%$#@&*(" etc.
Does it always need to contain a number?

source/reference/v2/payments-api/create-payment.rst
```diff
@@ -419,9 +419,19 @@ PayPal
    :type: string
    :condition: optional
 
-   If a description in the form ``Order <order number>`` is used, the order number is passed to PayPal as the
-   *invoice reference*. This field is searchable in the PayPal merchant dashboard. This field must be unique and should
-   not contain symbols. Alternatively, we will recognize the following keywords:
+   If a description in the form ``Order <order number>`` is used and the <order number> is right after the ``Order``
```
Timollie, line 422: Suggestion: Change to <order number> to <orderNumber>. Also might be nice to add that only the next value, separation by whitespaces, after "Order" is picked up.

Timollie (changes requested):

source/reference/v2/payments-api/create-payment.rst
ndijkstra: I think "symbols" is well known in the developer world so no need to specify all the symbols.

source/reference/v2/payments-api/create-payment.rst
ndijkstra: Do you have a suggestion for the whole sentence?

Timollie: Suggestion => If a description like ``Order <orderNumber>`` is used and the first value after ``Order``, separates by whitespaces, contains at least some numbers, it will be passed to PayPal as the *invoice reference*.

Not sure on "value", maybe use "entry" or "word". I do not know the right technical term for part of a string.

Edit example:
- Order Best Service ABS123 does not match as the first value after "Order", "Best", does not contain any numbers.

Then add example:
- Order ABS123 Best Service, will match and thus the first value after Order ``Order ABS123`` will be sent to PayPal

Timollie (changes requested):
See comment: https://github.com/mollie/api-documentation/pull/805/files#r714653907

Timollie (approved):

PR #806, opened 2021-09-14
html-collect-pages should be an iterable

waltertamboer:
When no redirects are configured, no iterable was returned.


source/extensions/sphinx_reredirects_fork.py
```diff
@@ -40,7 +40,7 @@ def setup(app: Sphinx):
 def init(app: Sphinx):
     if not app.config[OPTION_REDIRECTS]:
         logger.debug('No redirects configured')
-        return
+        return []
```
waltertamboer, line 43: Same solution as on line 51

willemstuursma (approved):

PR #819, opened 2021-11-02
Add new giftcards to the docs

MollieMateus:


MollieRick:
This PR is not needed since these are closed loop cards :-)

PR #820, opened 2021-11-05
Add optional landing_page parameter to authorize

ceeram:


source/changelog/v2/changelog.rst
```diff
@@ -3,6 +3,12 @@ Changelog
 Occasionally, we will add new resources, new fields, or new possible values to existing fields to the v2 Mollie API. All
 changes are documented here.
 
+November 2021
+=============
+Friday, 5th
+-----------
+- Added ``landing_page`` optional parameter to :doc:`Authorize </reference/oauth2/authorize>`.
```
vernondegoede, line 10:
```suggestion
- Added optional ``landing_page`` parameter to the :doc:`Authorize </reference/oauth2/authorize>` endpoint.
```

source/reference/oauth2/authorize.rst
```diff
@@ -75,6 +75,14 @@ Parameters
 
    Possible values: ``en_US`` ``nl_NL`` ``nl_BE`` ``fr_FR`` ``fr_BE`` ``de_DE`` ``es_ES`` ``it_IT``
 
+.. parameter:: landing_page
+   :type: string
+   :condition: optional
+
+   Allows you to redirect not logged in merchants to login or signup page. Defaults to login if omitted.
```
vernondegoede, line 82:
```suggestion
   Allows you to redirect unauthorized merchants to the login or signup pages at Mollie. Defaults to the login page if omitted.
```

dmvdbrugge: I wouldn't use "unauthorized" here, as that means something else in the context of an "authorize" flow. Trying to think of a suggestion.

dmvdbrugge: How about this:

> Allows you to specify if Mollie should show the login or the signup page, when the merchant is not logged in at Mollie. Defaults to the login page.

"Allows you to redirect" is incorrect because it's Mollie doing the redirecting, not them. Also remove the "if omitted" because it also ends up on the login page if we specify garbage.

ceeram: should i perhaps also remove login then as possible value? since anything that is not signup or omission will make it redirect to login

dmvdbrugge (approved):

PR #824, opened 2021-11-25
Mention 'Klarna Pay now' where relevant

mvdpanne:
Use the preferred order of:
- Klarna Pay now
- Klarna Pay later
- Klarna Slice it

### DIFF
--- a/source/reference/v2/refunds-api/create-order-refund.rst
+++ b/source/reference/v2/refunds-api/create-order-refund.rst
@@ -14,9 +14,6 @@ Create order refund
 
 When using the Orders API, refunds should be made for a specific order.
 
-When using *pay after delivery* payment methods such as *Klarna Pay later* and *Klarna Slice it*, your customer will
-receive a credit invoice with more information about the refunded products.
-
 If you want to refund arbitrary amounts, however, you can also use the
 :doc:`Create payment refund endpoint </reference/v2/refunds-api/create-payment-refund>` for Pay later and Slice it by
 creating a refund on the payment itself.
--- a/source/reference/v2/settlements-api/list-settlement-captures.rst
+++ b/source/reference/v2/settlements-api/list-settlement-captures.rst
@@ -14,9 +14,9 @@ List settlement captures
 
 Retrieve all captures in a certain settlement.
 
-Captures are used for *pay after delivery* payment methods. The only payment methods at the moment that have this flow
-are *Klarna Pay later* and *Klarna Slice it*. Captures are created when (part of) an Order is shipped. The capture is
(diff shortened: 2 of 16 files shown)


source/orders/overview.rst
```diff
@@ -20,8 +20,7 @@ This is mandatory for pay-after-delivery methods. Only shipped amounts will be s
 The following payment methods require the Orders API and cannot be used with the
 :doc:`Payments API </payments/accepting-payments>`:
 
-* *Pay after delivery* payment methods, such as Klarna Pay later and Klarna Slice it
-* Klarna Pay now
+* *Pay after delivery* payment methods, such as Klarna Pay now, Klarna Pay later and Klarna Slice it
```
mvdpanne, line 23: I wasn't entirely sure whether Klarna Pay now would be considered a 'pay after delivery' payment method.

In `source/reference/v2/settlements-api/list-settlement-captures.rst`, it is mentioned that:
> Captures are used for *pay after delivery* payment methods.

Which is why I decided to consider it as part of that group for this PR.

ndijkstra: No, Pay Now is not a pay after delivery method, but it is using captures

source/reference/v2/settlements-api/list-settlement-captures.rst
```diff
@@ -15,8 +15,8 @@ List settlement captures
 Retrieve all captures in a certain settlement.
 
 Captures are used for *pay after delivery* payment methods. The only payment methods at the moment that have this flow
```
ndijkstra, line 17: *pay after delivery* is mentioned here as well, but not the case for Pay now.

zoltanczirkos-mollie (approved):

source/reference/v2/refunds-api/create-order-refund.rst
```diff
@@ -14,8 +14,8 @@ Create order refund
 
 When using the Orders API, refunds should be made for a specific order.
 
-When using *pay after delivery* payment methods such as *Klarna Pay later* and *Klarna Slice it*, your customer will
-receive a credit invoice with more information about the refunded products.
+When using *pay after delivery* payment methods such as *Klarna Pay now*, *Klarna Pay later* and *Klarna Slice it*,
+your customer will receive a credit invoice with more information about the refunded products.
```
mvdpanne, line 18: @ndijkstra would the customer receive a credit invoice for Klarna Pay now as well?

mvdpanne: Does it make sense then to remove this paragraph?

PR #825, opened 2021-12-17
Add Apple Pay to first payment

VincentMollie:
Apple Pay is actually supported as a first payment, but the documentation doesn't include it.


VincentMollie:
@mvdpanne I found out that Apple Pay is actually supported when doing a first payment, can you please approve so the documentation is up-to-date?

mvdpanne (approved):

mvdpanne:
Nice find 👍

sandervanhooft:
Can you guys double check this? Because first it was available for recurring (at launch), then it was retracted, now it is available again?

PR #832, opened 2022-02-15
Explain how to upload files as part as an API call

willemstuursma:


sandervanhooft:
Good to see you online again @willemstuursma !

Is there any other reference to the usage of https://api.mollie.com/v2/organizations/me/documents ?

willemstuursma:
Hi @sandervanhooft. 

At the moment we don't have any public API endpoints that support file uploads. I also cannot give you any more information on the endpoint mentioned at the moment.

mollierobbert:
I assume this is a draft MR, then, that does not have to be reviewed until we launch the concerning endpoint?

willemstuursma:
Hi @mollierobbert Thanks for reaching out. I lost track of this MR. 

At the moment there is no public endpoint that supports uploading files. However, I wanted to write this guide to set a norm on how to do uploads in the v2 API as well as support engineers implementing internal API endpoints in our various internal clients. 

Hence, a review would be appreciated with a merge as a goal. What do you think?

mollierobbert:
Ah, in that case maybe we can make it an internal 'API RFC' first? I'll share the new guidelines.

PR #837, opened 2022-03-03
PC-255 Add `en_GB` as accepted locale for all APIs

dabroek:


mvdpanne:
Since the new locale is added to the v1 API as well, would it make sense to add an item to the v1 changelog as well?

mvdpanne (approved):

PR #838, opened 2022-03-04
Pietro/api keys over access codes and chargeback info

pgoodjohn:


source/connect/refunds-and-chargebacks.rst
```diff
@@ -148,3 +148,21 @@ funds that were sent to connected accounts ``org_23456`` and ``org_56789``.
        ]
        "...": { }
    }
+
+Chargebacks of Split Payments
+----------------------------------
+
+Whenever one of your split payment gets charged back, your Platform will be responsible for the Mollie fees and for the initial
```
antodippo, line 155: Here I don't understand the "...your Platform...", maybe it should be just "you"?

source/connect/refunds-and-chargebacks.rst
```diff
@@ -148,3 +148,21 @@ funds that were sent to connected accounts ``org_23456`` and ``org_56789``.
...
+submerchants, you might be eligible to receive a compensation for the amount that was routed to the other accounts.
+
+Specifically, if the payment was split between you and only one other organization you will be automatically compensated for the
+amount that was routed to the submerchant, independently of the amount of the chargeback but limited to the originally routed amount. 
+
+If the payment was split across multiple submerchants, you will only receive compensations for each of the routes if the chargeback was for 
+the full amount of the original payment (or higher). If we receive a chargeback for a lower amount than the original payment, we will
+detract the amount from your balance and you will not receive any compensation for it.
+
+In case you and your submerchant(s) decide to object to the chargeback and can provide enough evidence for it to be reversed, any
```
antodippo, line 166: It's not clear to me how does this would work: that would mean that for a payment split to SubmA and SubmB that is charged back, if SubmA provides evidence we reverse that specific route's compensation? Are we able to do that? :D

source/connect/refunds-and-chargebacks.rst
pgoodjohn: I saw we say "your Platform" for the Refunds section as well so I wanted to keep it in line! Maybe it is better to change  them both however 🤔

antodippo (approved):

source/connect/refunds-and-chargebacks.rst
sandervanhooft: How about "your Mollie account will be charged..."?

pgoodjohn: @sandervanhooft probably best 😁 I'll update it in both sentences.

PR #840, opened 2022-03-28
Updated Node.js examples for 3.6.0

Pimm:
* Updated examples to use dromedary case (`customerPayments` instead of `customers_payments`). Dromedary case is preferred as it is standard in the JavaScript community, plus snake case will be removed in 4.0.0. 🐪
* Updated examples to use the iteration API instead of the pagination API. The iteration API is likely what most users will want to use.
* Removed the asynchronous IIFE: `(async () => { … })()`. Promises are well-established in the JavaScript community today. I don't suspect the top-level `await` keyword would confuse anyone. (In fact, at this point I think it's more likely that the asynchronous IIFE confuses users.)
* Fixed a mistake in the example on [the page about user agent strings](//docs.mollie.com/integration-partners/user-agent-strings).

### DIFF
--- a/source/components/overview.rst
+++ b/source/components/overview.rst
@@ -262,19 +262,17 @@ Example
       const { createMollieClient } = require('@mollie/api-client');
       const mollieClient = createMollieClient({ apiKey: 'live_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM' });
 
-      (async () => {
-        const payment = await mollieClient.payments.create({
-          method: 'creditcard',
-          amount: {
-            currency: 'EUR',
-            value: '10.00', // We enforce the correct number of decimals through strings
-          },
-          description: 'Order #12345',
-          redirectUrl: 'https://webshop.example.org/order/12345/',
-          webhookUrl: 'https://webshop.example.org/payments/webhook/',
-          cardToken: 'tkn_UqAvArS3gw'
-        });
-      })();
+      const payment = await mollieClient.payments.create({
+        method: 'creditcard',
+        amount: {
+          currency: 'EUR',
+          value: '10.00'
+        },
+        description: 'Order #12345',
+        redirectUrl: 'https://webshop.example.org/order/12345/',
+        webhookUrl: 'https://webshop.example.org/payments/webhook/',
+        cardToken: 'tkn_UqAvArS3gw'
+      });
 
 Response
 ^^^^^^^^
--- a/source/reference/v2/refunds-api/cancel-payment-refund.rst
+++ b/source/reference/v2/refunds-api/cancel-payment-refund.rst
@@ -76,9 +76,9 @@ Example
       const { createMollieClient } = require('@mollie/api-client');
       const mollieClient = createMollieClient({ apiKey: 'test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM' });
 
-      (async () => {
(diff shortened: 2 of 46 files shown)


source/components/overview.rst
```diff
@@ -262,19 +262,17 @@ Example
...
-            currency: 'EUR',
-            value: '10.00', // We enforce the correct number of decimals through strings
-          },
-          description: 'Order #12345',
-          redirectUrl: 'https://webshop.example.org/order/12345/',
-          webhookUrl: 'https://webshop.example.org/payments/webhook/',
-          cardToken: 'tkn_UqAvArS3gw'
-        });
-      })();
+      const payment = await mollieClient.payments.create({
```
daveymoores, line 265: I assume `@mollie/api-client` has a supported node version that is greater than the one in the `.nvmrc` in this project (12.x.x), but just to flag that top-level await is only supported in node v13.3 and above

source/reference/v2/refunds-api/cancel-payment-refund.rst
```diff
@@ -76,9 +76,9 @@ Example
       const { createMollieClient } = require('@mollie/api-client');
       const mollieClient = createMollieClient({ apiKey: 'test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM' });
 
-      (async () => {
-        const status = await mollieClient.payments_refunds.cancel('re_4qqhO89gsT', { paymentId: 'tr_WDqYK6vllg' });
-      })();
+      await mollieClient.paymentRefunds.cancel('re_4qqhO89gsT', {
```
daveymoores, line 79: Should this be returning a value?

daveymoores (approved):

source/reference/v2/refunds-api/cancel-payment-refund.rst
Pimm: It returns `true`, for historic reasons. I personally feel putting the `true` value in a variable like that is more confusing than helpful, considering how an error is thrown if it fails (rather than returning `false`). Let me know if you disagree.

source/components/overview.rst
Pimm: The current version supports Node.js 6.14.×, and the new version (4.0.0) will support Node.js 8.×. So there is no top-level `await` guarantee.

I've given this a bit of thought myself. The thing is, the code example:
```javascript
const { createMollieClient } = require('@mollie/api-client');
const mollieClient = createMollieClient({ apiKey: '[…]' });

const payment = await mollieClient.payments.create({
```
is very incomplete. In reality, if you use this library with Express, for instance, you might create the client and put it in `app.locals` but create the payment in a handler. In other words, calling `createMollieClient` and `payments.create` in such quick succession is unnatural anyway.

In my mind, the goal of the example is to demonstrate the steps required to create a payment:

1. Import library
2. Create client
3. Create payment

In that context, I feel the IIFE just adds noise. It's only needed if you want to copy this example into the top level and then execute it in an old Node.js version, which sounds like a very specific and atypical scenario to me.

Let me know what you think. I'm on Slack.

daveymoores: I agree with this. I made the comment as I assumed that was what the IIFE was trying to convey. All good.

Pimm:
Thanks, Daddy V! 💕

PR #852, opened 2022-05-20
Document the in3 payment method

mvdpanne:
- List in3 on the documentation overview page.
- List in3 and Klarna Pay now on the 'Build your own checkout' page.
- List the expiry time of in3 payments.
- List in3 as a supported payment method for the Orders API.
- List in3 as a supported payment method for retrieving a payment.
- Add an entry in the v2 API changelog.

To do/to be confirmed:
- [x] Is the URL that I used on the overview page correct?
- [x] Update the changelog to feature the correct date once we are ready to ship this.

### DIFF
--- a/source/changelog/v2/changelog.rst
+++ b/source/changelog/v2/changelog.rst
@@ -3,6 +3,12 @@ Changelog
 Occasionally, we will add new resources, new fields, or new possible values to existing fields to the v2 Mollie API. All
 changes are documented here.
 
+June 2022
+=========
+????, ??
+--------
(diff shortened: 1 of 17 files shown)


mollierobbert (approved):

zoltanczirkos-mollie:
In3 could be mentioned here as well: `/orders/why-use-orders.rst`, in the list item starting with 'Pay after delivery payment methods'. Although that list might not be meant to be a comprehensive list of all Orders-API-only payment methods.

mvdpanne:
Thanks Zoltán! It triggered me to add in3 to some more relevant pages and make some changes regarding our use of 'pay after delivery payment methods' in the documentation.

PR #858, opened 2022-06-22
Added external link to Postman Collection to reference sidebar

Jonas4Mollie:
This is a link to the public Github repo containing the Postman Collections that hold all our API references.

We decided to add it to the API Reference part as that is the scope of the collections. Then, in order to respect the sidebar structure, we decided to add a new "header" that we titled "Additional Resources" (and placed it at the end) with only one list entry called "Postman Collections" (plural as there are 2 of them).

We removed all class-references that are triggered when a page is active (highlighting it) as this is an external link, in contrast to all the others in the sidebar.

Up to discussion: should we set target="_blank" on this link as it is an external one?


MollieRick:
@Jonas4Mollie Hi! I think we should set `target="_blank"` for sure. However, I am a bit doubting about the possible where you put the link. Are these collections v2 only?

Jonas4Mollie:
@MollieRick both collections do only support the v2 API, so I tried to get the menu item listed only when the v2 API is selected.
Isn't the web framework handling the target of the external link by default, by the way?

MollieRick:
No it doesn't :D 

What do you think @pgoodjohn ?

pgoodjohn:
I don't think it should be in the API reference to be honest. That should just be a list of endpoint and their details. I think it should be somewhere in the Overview section of the docs

Jonas4Mollie:
so how will we decide on the way forward? Do we need a meeting?
My attention was to first make it available and later improve it.

Jonas4Mollie:
Can we agree on this approach: I will update it to open in a new tab/window and then we proceed and first publish and later review?

PR #863, opened 2022-07-13
Update PR for adding the Postman Collections to the Mollie Docs

Jonas4Mollie:
This PR contains the second attempt to publish the Postman Collections in the Mollie Docs. This time they are added to the Overview sidebar instead of the Reference sidebar and the hyperlink should open in a new window as it is an external link.
Let me know if something should be changed!


mollierobbert:
Shipped this in a slightly different way. Closing the issue.

PR #870, opened 2022-08-31
Add hint to use different profiles for different Apple Pay domains

fjbender:
We should make it more clear in the documentation that Profiles and Domains are coupled together. Carelessly using the same profile for stage and live systems might result in payment failures.


willemstuursma:
I'd like to check up on this tomorrow, I'm not sure this holds.

fjbender:
Assumption is not entirely true, the hint is obsolete.

PR #872, opened 2022-09-19
Clarify webhookUrl behavior in subscriptions

fjbender:


mollierobbert:
Thanks Flo, I'll add this to the new docs site.

PR #873, opened 2022-09-27
Add the read-only "archived" property to payment links

ArtemGordinsky:
As-per the RFC, Payment Links' `archived` property is now exposed for reading.

Hopefully, my description makes sense.

Screen Shot 2022-09-27 at 11 37 22


mollierobbert (approved):

mollierobbert:
'Unavailable for payment' is perhaps a bit vague. Can you change it to something like:

'Whether the payment link is archived. Customers will not be able to complete payments on archived payment links.'

PR #881, opened 2022-10-25
Update paypal-express-checkout-button.rst

SalimAtMollie:
This improves the PayPal express documentation on the propagation of shipping and billing addresses.

TECHSUP-3533

### DIFF
--- a/source/orders/paypal-express-checkout-button.rst
+++ b/source/orders/paypal-express-checkout-button.rst
@@ -28,7 +28,7 @@ some small tweaks. You can use the following steps as a guideline for the basic 
 #. Place the button on the intended page, like a product page.
 
 #. Your website :doc:`creates an order </reference/v2/orders-api/create-order>` via the Orders API after the consumer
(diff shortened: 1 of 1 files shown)


source/orders/paypal-express-checkout-button.rst
```diff
@@ -38,11 +38,16 @@ some small tweaks. You can use the following steps as a guideline for the basic
                 all, your request will result in an HTTP status code ``422 Unprocessable Entity``.
 
 #. Redirect your customer to the ``checkout`` URL which you can find in the response of the Create Order
-   API. The customer will select the address where the product(s) needs to be send to and completes the
+   API. The customer will select the address where the product(s) needs to be sent to and complete the
```
mollierobbert, line 41: The customer completes... that one was correct, I think. Maybe break up the sentence in two?

> The customer will select the address where the product(s) need to be sent to. Then, the customer can complete the payment.

SalimAtMollie: Yep! Definitely agree that we should probably just break up the sentence in two.

mollierobbert (approved):

PR #889, opened 2022-11-22
Php update calls

sandervanhooft:
The old recommended approach was to do update calls like this with mollie-api-php:

```php
$payment = $mollie->payments->get($id);
$payment->webhookUrl = "https://updated.com";
$payment->update();
```

This way however, all updatable fields are sent over to Mollie's API on each update request. I have encountered multiple tickets over time from integrators asking for more granularity and control here. So a while ago I've added methods to directly make the update call and pass in all id's and payload directly.

```php
$payment = $mollie->payments->update($id, [
  "webhookUrl" => "https://updated.com",
]);
```

While I feel the old examples look nicer, the way described below has been now proven to result in a better developer experience. So I've matched the examples accordingly. :)


sandervanhooft:
cc @Sebastian-mollie

PR #898, opened 2022-12-22
Add canceled refund status

dabroek:
Add canceled status to list of refund statuses

### DIFF
--- a/source/payments/refunds.rst
+++ b/source/payments/refunds.rst
@@ -35,7 +35,7 @@ Refunds have their own status, independent of the Payment or Order they were cre
   The refund is being processed. Cancelation is no longer possible.
 
 *refunded*
-  The refund has been completed and your customer has received the funds.
+  The refund has been completed and your customer has either received the funds or the funds are on their way.
 
 *failed*
   The refund has failed after processing. For example, the customer has closed his / her bank account. The funds will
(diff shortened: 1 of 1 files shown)


source/payments/refunds.rst
```diff
@@ -32,7 +32,7 @@ Refunds have their own status, independent of the Payment or Order they were cre
   The refund is ready to be sent to the bank. You can still cancel the refund if you like.
 
 *processing*
-  The refund is being processed. Cancellation is no longer possible.
+  The refund is being processed. Cancelation is no longer possible.
 
 *refunded*
   The refund has been settled to your customer.
```
sandervanhooft, line 38: A bit in doubt here. Is this appropriate English?

`The refund has been settled to your customer.`

dabroek: Correct would be "The refund is settled with your customer."

More comprehensible could be: "The refund has been completed and your customer has received it."

dabroek: FYI. Added the disclaimer that "the funds can still be on their way."

sandervanhooft: Even better!

sandervanhooft (approved):

mollierobbert (approved):

PR #914, opened 2023-03-07
Note on updating order billing address: not propagated to supplier

zoltanczirkos-mollie:
Add note on updating order billing address: it is not propagated to the supplier if it doesn't support it.


pgoodjohn (approved):

sandervanhooft:
Ok, but what does this practically mean if you're integrating this?

zoltanczirkos-mollie:
I'll share some details over slack

PR #916, opened 2023-03-14
Rework hosted onboarding guide

zsoltparal-mollie:
This PR changes the hosted onboarding guide, clarifying a few points on how to set up and monitor the onboarding integration. It also expands on a few points, adding recommendations to make the onboarding experience even smoother.


source/connect/onboarding.rst
```diff
@@ -41,10 +30,33 @@ To onboard customers, you commonly need these three permissions:
        Onboarding API
      - Submit onboarding data on behalf of the merchant.
 
-Step 3: Have your customer sign up and grant permission
--------------------------------------------------------
-Now that have an app with the OAuth authorization flow implemented and the right permissions configured, you can start
-sending your customers to the authorization screen.
+   * - ``organizations.write``
```
pgoodjohn, line 33: Shouldn't we also have `organization.read` here?

source/connect/onboarding.rst
```diff
@@ -53,20 +65,33 @@ Once they are logged in, your customers will be asked to give permission to your
...
+.. note:: This is possible as long as the onboarding status of the new account is set to ``needs-data``.
 
-Before you do so, you may want to provide onboarding data you already have on the customer using the
-:doc:`Submit onboarding data </reference/v2/onboarding-api/submit-onboarding-data>` endpoint. This is possible as long
-as the onboarding status of the new account is set to ``needs-data``.
+You can use the :doc:`Get onboarding status </reference/v2/onboarding-api/get-onboarding-status>` endpoint to get a link
+to the white-labeled onboarding flow. We recommend that you redirect your customer right away to this URL so they can
+provide the required onboarding information.
+
+.. code-block-selector::
```
pgoodjohn, line 83: I would add an example response with the hosted onboarding url rather than this PHP code example here since in the [api reference](https://docs.mollie.com/reference/v2/onboarding-api/get-onboarding-status) we don't show it

pgoodjohn (changes requested):

source/connect/onboarding.rst
zsoltparal-mollie: You are right, I added it.

source/connect/onboarding.rst
zsoltparal-mollie: I changed the code block to include an example response instead and added a reference to it in the text.

pgoodjohn (approved):

PR #919, opened 2023-03-17
Refund webhooks are also triggered for status `processing`

fjbender:


dabroek:
Your commit message is a little misleading, because the webhook is triggered upon initiating the refund process, which can then result in the refund having the state processing, refunded, or failed. It is not "triggered for the status processing". But that's just semantics. ;-)

PR #920, opened 2023-03-28
Update multicurrency.rst

hwysoszynski:
To highlight that we are offering Apple Pay also for these currencies, I would like to propose adding them to add the payment method list too.


mollierobbert:
Good call. I added this to the new docs site.

PR #921, opened 2023-03-29
Fix sphinx does not work when Python's user base path is not part of $PATH

willemstuursma:
Various `make` commands such as `make html-reload` did not work on my local machine.

Turned out that the user path where the dependencies are installed by `pip` was not in my `$PATH`. 

A little `Makefile` magic takes care of this. 

In addition, @ZsoltMollie told me the `"node-sass` module isn't needed. One of its dependencies requires python 2 which makes everything more complicated, so good riddance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,12 @@
 
 # Minimal makefile for Sphinx documentation
 #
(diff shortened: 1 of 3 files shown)


mollierobbert:
LGTM, thanks Willem!

PR #936, opened 2023-05-17
Update Python and NPM dependencies

zsoltparal-mollie:
This MR updates all NPM and Python dependencies to the newest version that we can support. The two notable exceptions are:

1. ESLint, as the configuration has ^7.0.0 as the peer dependency, so I kept it in that range instead of going for v8
2. Sphinx 7.0.0, which included a lot more breaking changes than 6.0.0, so I went with the easier upgrade first


source/reference/v2/subscriptions-api/update-subscription.rst
```diff
@@ -215,9 +215,6 @@ Response
 .. code-block:: json
    :linenos:
 
-    HTTP/1.1 200 OK
-    Content-Type: application/hal+json
```
zsoltparal-mollie, line 219: This file failed to build with `code-block: http` for some reason. All other code blocks worked, but this one would keep failing.

requirements.txt
```diff
@@ -1,5 +1,5 @@
-sphinx==1.8.6
+sphinx==6.2.1
 jinja2==3.0.0
+sphinx-autobuild==2021.3.14
 urllib3==1.*
```
mvdpanne, line 4: NBC: I think it is likely that we now no longer need to pin to version 1.x of this package. This was related to being 'stuck' on such an old version of sphinx. The same counts for the `jinja2` package.

zsoltparal-mollie: I tried removing these, but the build fails without them.

mvdpanne: Too bad, thanks for trying, though! 😄

zsoltparal-mollie:
It seems like our Sphinx theme is not compatible with the latest version and it seems to be unmaintained. We should think about how to proceed.

PR #940, opened 2023-05-25
Made canceledAt in the get subscription endpoint optional

Pimm:
The `canceledAt` parameter is optional according to the description for it:
 > This parameter is omitted if the payment is not canceled (yet).

([Thanks](//github.com/mollie/mollie-api-node/pull/317) to @janpaepke)


mollierobbert:
Good find. Should be corrected now in the new Mollie docs.

PR #941, opened 2023-05-29
ENG-597 - Disallow over-refunding by default

vernondegoede:
Mollie no longer allows over-refunding by default. This is only available when configured on the org-level by a Mollie employee, but since this is an edge-case now.


sandervanhooft:
@vernondegoede potential BC?

mollierobbert (approved):

PR #944, opened 2023-06-09
Improvements for POS docs

cristinaachimmollie:


source/reference/v2/terminals-api/get-terminal.rst
```diff
@@ -48,19 +49,27 @@ Response
...
+   for you, it will be move into this state.
+
+   ``inactive`` means that the terminal has been deactivated, which can mean that you returned the device to us,
+   or you requested to move it to another profile or organization.
 
 .. parameter:: brand
    :type: string
 
-   The brand of the terminal.
+   The brand of the terminal. In most of the cases the value of this field will be "PAX".
```
mollierobbert, line 66: Maybe just make it:

> The brand of the terminal. For example, 'PAX'.

source/reference/v2/terminals-api/get-terminal.rst
```diff
@@ -48,19 +49,27 @@ Response
...
 
 .. parameter:: model
    :type: string
 
-   The model of the terminal.
+   The model of the terminal. The model can differ, in most of the case this field's value will be "A920",
```
mollierobbert, line 71: Same as above, maybe an example is better and then we don't have to amend the docs when new models come out.

> The model of the terminal. For example for a PAX A920, this field's value will be 'A920'.

mollierobbert (approved):
Left some non-blocking suggestions.

mollierobbert (approved):

PR #950, opened 2023-06-22
Refresh Application Fee documentation

pgoodjohn:


source/connect/application-fees.rst
```diff
@@ -1,32 +1,27 @@
 Receiving application fees
 ==========================
-An easy way to split a payment between your platform and your connected merchant accounts is to charge
-**Application fees** on the incoming payments on the connected accounts.
+An easy way to earn on a transaction is by charging an **Application fee** on any  incoming payments of a connected account.
 
-This solution is ideal for simpler use cases, where the connected account remains fully liable and in control of the
-payment, and your platform only deducts a fee.
+In this set-up the connected account pays their own Mollie payment fees, receives a Mollie invoice, and remains liable for potential 
+refunds and chargebacks. If you want to split a payment with another party (for instance if you are a marketplace),
```
mollierobbert, line 6: I'd use commas here. Parentheses are not commonly used for sub-sentences in English.

> ...with another party, for instance if you run a marketplace, we offer...

source/connect/application-fees.rst
```diff
@@ -1,32 +1,27 @@
...
 
-The connected account in this case will still have their own dashboard, pay their own Mollie payment fees, receive a
-Mollie invoice, and remain liable for potential chargebacks.
+.. admonition::
+      **Example**
 
-An example use case is a ticketing platform that charges a fee per successful payment to each connected theater. In this
-case, each theater has an account with Mollie, and they have the ticketing platform's OAuth app authorized to create
-payments on their behalf.
+      A ticketing platform wants to charge a fee per successful payment to each connected theatre. In this case, each theatre has an account
```
mollierobbert, line 12: We use American English, so 'theater'.

source/connect/application-fees.rst
```diff
@@ -1,32 +1,27 @@
...
-The ticketing platform can add an application fee to each payment. When the payment is successful, the fee specified in
-the application fee is transferred from the theatre's account to the platforms account.
-
-Mollie will then collect and settle the application fees to the ticketing platform. The ticketing platform itself is
-responsible for separately invoicing the theaters for the incurred fees, and for handling VAT.
-
-For more advanced use cases, for example if you want to cover the Mollie payment fees yourself, or for example if you
-want to split a payment with another party, we offer :doc:`Split payments </connect/splitting-payments>`.
+      The ticketing platform can add an application fee to each payment. When the payment is successful, the fee specified in the application
+      fee is automatically transferred from the theatre’s account to the platforms account.
```
mollierobbert, line 16: > ...to the platform's account.

source/connect/application-fees.rst
```diff
@@ -75,32 +70,44 @@ They are created by passing additional parameters to the
...
-| your account manager at Mollie.
-
-Recurring
--------------
-Application fees are both supported on recurring payment and on subscriptions.
+*Payments API*
+Any application fee has a maximum as sufficient funds need to be available to pay the payment fee. 
+The maximum is related to the payment amount and the maximum can be calculated as following: 
+
+Max application fee = amount of the payment - 1.21 × (0.29 + (0.05 × the amount of the payment))
```
mollierobbert, line 90: Can we not simplify this by now? It's annoying to deal with this unpredictable maximum, and it's inconsistent with the Orders API.

source/connect/application-fees.rst
```diff
@@ -75,32 +70,44 @@ They are created by passing additional parameters to the
...
-|
-| The maximum application fee per payment is 10% of the total amount, up to a maximum of €2.00. If a higher maximum is
-| required for your business, you can request this via Mollie's `customer service <https://www.mollie.com/contact>`_ or
-| your account manager at Mollie.
-
-Recurring
--------------
-Application fees are both supported on recurring payment and on subscriptions.
+*Payments API*
+Any application fee has a maximum as sufficient funds need to be available to pay the payment fee.
```
mollierobbert, line 87: How about:

> Since Mollie usually also charges a fee on every payment, there is a limit to the maximum application fee amount.
>
> The Mollie fee charged will depend on which method the consumer selects. To keep things simple however, we reserve €0,29 + 5% of the payment amount for us to be able to charge the Mollie fee, multiplied by 21% to cover Dutch VAT.
>
> This means the maximum can be calculated as follows:

source/connect/application-fees.rst
```diff
@@ -75,32 +70,44 @@ They are created by passing additional parameters to the
 
 Testing application fees
 ------------------------
-Application fees work in test mode as well.
+You can submit the `applicationFee` parameter in test mode. The application fee will be visible in the payment details
```
mollierobbert, line 73: Single backtick makes it cursive I believe. Double backticks all the way.

mollierobbert (changes requested):
See comments.

source/connect/application-fees.rst
pgoodjohn: It is not unpredictable though, the formula is always the same with the payment amount you can reliably calculate what the maximum will be. 

For now I would keep these docs as they are to be as transparent as possible, with the extended description we added I feel like it is easier to understand what's going on in the formula.

For the orders limit, I would like to unify or completely remove these maxima, but that's something that I'd prefer to approach with more thought rather than just yoloing it while changing some docs 😛

source/connect/application-fees.rst
```diff
@@ -91,18 +91,18 @@ Maximum application fees
 Since Mollie usually also charges a fee on every payment, there is a limit to the maximum application fee amount.
 
 The Mollie fee charged will depend on which method the consumer selects. To keep things simple however, we reserve 
-€0,29 + 5% of the payment amount for us to be able to charge the Mollie fee, multiplied by 21% to cover Dutch VAT.
+€0,35 + 6% of the payment amount for us to be able to charge the Mollie fee.
 
 This means the maximum can be calculated as follows:
 
-Max application fee = amount of the payment - 1.21 × (0.29 + (0.05 × the amount of the payment))
+Max application fee = amount of the payment - 0.35 + (0.06 × the amount of the payment)
```
mollierobbert, line 98: I believe you rounded to €0,36 in code.

mollierobbert (approved):

PR #987, opened 2024-03-20
Update status-changes webhook info

DavidDeSloovere:
Missing webhook info for autorized.
Found the info at https://docs.mollie.com/overview/webhooks

### DIFF
--- a/source/payments/status-changes.rst
+++ b/source/payments/status-changes.rst
@@ -21,7 +21,7 @@ them. Then we will show you how the statuses are connected.
 
 ``authorized``
     If the payment method supports captures, the payment method will have this status for as long as new captures can be
-    created.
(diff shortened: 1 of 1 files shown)


DavidDeSloovere:
I'm not a fan that the copy is different for every status. Can I update this also?

joosttk:
Thanks for your PR @DavidDeSloovere. One comment on this, is that we _will_ call a webhook when the payment is not part of an Order. Currently, we only support this for credit card payments that leverage our function to [place a hold on a card](https://docs.mollie.com/payments/place-a-hold-for-a-payment).

But if your payment is part of an Order, we will call the Order webhook (and in that case indeed we will not call a Payment webhook). For more information, see our [Webhooks page](https://docs.mollie.com/overview/webhooks).

PR #992, opened 2024-04-11
Fix Mollie website partners link in all docs

SalimAtMollie:
Changed all occurances of https://www.mollie.com/partners to https://www.mollie.com/nl/partner-program/saas-ecommerce#form


joosttk:
The new URL's now all have `nl` in them, which means they will always direct to the Dutch page. You can remove this part.

joosttk (approved):

PR #1007, opened 2024-05-30
docs: remove duplicate word

luizdamim:
The word `will` is duplicated in https://docs.mollie.com/reference/v2/payments-api/create-payment#payment-method-specific-parameters

image


mollierobbert:
Thanks Luiz, fixed this in the new Mollie docs.

PR #1008, opened 2024-06-04
Joosttk/PC-1465-addresses-payments-api

joosttk:
Adds billingAddress and shippingAddress to Payments API.


sandervanhooft:
It's also worth mentioning here that the `Payment.billingEmail` and `details.shippingAddress` are being marked for deprecation.

